PR #1, opened 2020-06-23
Spring boot setup

jsheng112:
Setup a basic hello world application for spring boot.


capstone/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+
+
+      <!-- Provides `mvn package appengine:run` for local testing
+           and `mvn package appengine:deploy` for deploying. -->
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>appengine-maven-plugin</artifactId>
+        <version>2.2.0</version>
+        <configuration>
+          <deploy.projectId>elfel-spring</deploy.projectId>
```
elefeint, line 57: Bwahahaha! I am in your project messing up your deploys.

capstone/src/test/java/com/step902020/capstone/CapstoneApplicationTests.java
```diff
@@ -0,0 +1,13 @@
+// package com.step902020.capstone;
```
elefeint, line 1: Out of curiosity, does the testd work if uncommented? It seems like it should, although it uses JUnit5, and you've only used JUnit4 so far.

jsheng112: It throws errors when uncommented.

PR #8, opened 2020-06-26
fetch profile page and update user profile

jsheng112:
Added basic HTML/CSS to support a user profile page and form. Can now fetch user information and fill in the user profile from Datastore based on whether the user is an individual or is an organization. Form now supports editing user profile information and storing it in Datastore. 

Authentication is complete but for the convenience of running locally the email is preset if it fails to get the email from authentication.


capstone/src/main/java/com/step902020/capstone/User.java
```diff
@@ -0,0 +1,110 @@
...
+  @Field(name="lastname")
+  String lastName;
+
+  @Field(name="email")
+  String email;
+
+  @Field(name="university")
+  String university;
+
+  @Field(name="description")
```
lbourret, line 27: What's the description for? Like students can add a bio for themselves?

capstone/src/main/java/com/step902020/capstone/User.java
```diff
@@ -0,0 +1,110 @@
...
+    this.description = description;
+    this.image = image;
+    // savedEvents = new HashSet<String>();
+    // savedEvents.add("event1");
+    // savedEvents.add("event2");
+    // savedEvents.add("event3");
+    // savedOrganizations = new HashSet<String>();
+    // savedOrganizations.add("ACM"); 
+    // savedOrganizations.add("ACM2");
+    // savedOrganizations.add("ACM3");
```
lbourret, line 56: ! good idea the to-dos are really helpful to understand your next steps

capstone/src/main/java/com/step902020/capstone/UserServlet.java
```diff
@@ -0,0 +1,47 @@
...
+  public RedirectView saveUser(
+      @RequestParam("firstname") String firstname,
+      @RequestParam("lastname") String lastname, 
+      @RequestParam("email") String email, 
+      @RequestParam("user-type") String userType,
+      @RequestParam("university") String university,
+      @RequestParam("description") String description) throws IOException {
+
+    // first delete the old entry the add the new one
+    this.userRepository.deleteByEmail(email);
```
lbourret, line 40: Is a way to update a user's profile?

capstone/src/main/resources/static/profile.html
```diff
@@ -0,0 +1,68 @@
...
+            <li><a href="about.html">Find Events</a></li>
+            <li><a href="projects.html">Find Organizations</a></li>
+            <li><a href="experiences.html">My Account</a></li>
+            <li><a href="gallery.html">About Us</a></li>
+        </ul>
+    </nav>
+    <br/>
+    <div class="left">
+        <ul>
+            <li class="active"><a href="index.html">My Profile</a></li>
```
lbourret, line 22: index.html -> profile.html

capstone/src/main/resources/static/savedevents.html
```diff
@@ -0,0 +1,41 @@
...
+<head>
+    <meta>
+    <title>My Account</title>
+    <link rel="stylesheet" href="style.css">
+    <script src="script.js"></script>
+</head>
+<body onload="getUser();">
+    <nav>
+        <ul>
+            <li class="active"><a href="index.html">College Activity Finder</a></li>
```
lbourret, line 12: update links

capstone/src/main/resources/static/savedevents.html
```diff
@@ -0,0 +1,41 @@
...
+            <li><a href="about.html">Find Events</a></li>
+            <li><a href="projects.html">Find Organizations</a></li>
+            <li><a href="experiences.html">My Account</a></li>
+            <li><a href="gallery.html">About Us</a></li>
+        </ul>
+    </nav>
+    <br/>
+    <div class="left">
+        <ul>
+            <li><a href="index.html">My Profile</a></li>
```
lbourret, line 22: index.html -> profile.html

capstone/src/main/resources/static/style.css
```diff
@@ -0,0 +1,133 @@
+body {
+  background-color: #aed1d6;
```
lbourret, line 2: wait are we making all the pages this color or will the event page be the dark color schema.

jsheng112: No its gonna be the dark theme, I just haven't done any css formating yet.

capstone/src/main/java/com/step902020/capstone/User.java
```diff
@@ -0,0 +1,110 @@
...
+  @Field(name="image")
+  String image;
+
+// these fields are going to be implemented in the future.
+
+//   @Field(name="saved-events")
+//   HashSet<String> savedEvents;
+
+//   @Field(name="saved-organizations")
+//   HashSet<String> savedOrganizations;
```
oasemota99, line 39: I agree with Lian, the TODOs are incredibly helpful! But I know the one thing Elena always gets on me for are comments, she may want you to stash them for later and just to keep running code in PRs. Just a heads up!

capstone/src/main/resources/static/profile.html
```diff
@@ -0,0 +1,68 @@
...
+            <label>First Name</label>
+            <input type="text" name="firstname" placeholder="Your first name..">
+            <label>Last Name</label>
+            <input type="text" name="lastname" placeholder="Your last name..">
+            <label>Your Email</label>
+            <input type="text" name="email" id="email-form">
+            <label>Are you an individual or a club/organization?</label>
+            <select name="user-type">
+                <option value="individual">Individual</option>
+                <option value="organization">Club or Organization</option>
```
oasemota99, line 58: Are you planning on creating separate entities for individuals and organizations? I know Mike mentioned it earlier. I may have missed the memo lol, but if you are, we may want to make separate forms for the two

capstone/src/main/resources/static/profile.html
```diff
@@ -0,0 +1,68 @@
...
+    <script src="script.js"></script>
+</head>
+<body onload="getUser();">
+    <nav>
+        <ul>
+            <li class="active"><a href="index.html">College Activity Finder</a></li>
+            <li><a href="about.html">Find Events</a></li>
+            <li><a href="projects.html">Find Organizations</a></li>
+            <li><a href="experiences.html">My Account</a></li>
+            <li><a href="gallery.html">About Us</a></li>
```
oasemota99, line 16: You should change the file names for each function it's going to be used for

capstone/src/main/java/com/step902020/capstone/UserServlet.java
jsheng112: You are right this is probably not the best way to do updates. I will investigate into this more.

capstone/src/main/resources/static/profile.html
jsheng112: Yes I am making separate forms for individuals and organizations. I think the way I am going to implement this is to have a toggle selection that switches between the two forms.

capstone/pom.xml
```diff
@@ -36,6 +36,33 @@
       <artifactId>spring-cloud-gcp-starter-data-datastore</artifactId>
       <version>1.2.3.RELEASE</version>
     </dependency>
+
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-api-1.0-sdk</artifactId>
```
meltsufin, line 42: Are you using any of the AppEngine APIs? Since we're not using the Users API, I don't think so.
Spring Cloud GCP should have everything you'll need.

capstone/pom.xml
```diff
@@ -36,6 +36,33 @@
...
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-test</artifactId>
+      <version>2.2.2.RELEASE</version>
```
meltsufin, line 56: scope test

capstone/src/main/java/com/step902020/capstone/UserServlet.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.web.servlet.view.RedirectView;
+import com.google.gson.Gson;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import java.io.IOException;
+
+
+/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+@RestController
+public class UserServlet {
```
meltsufin, line 20: Since we're in the Spring Boot world now, you should probably call the class `UserController`.

capstone/src/main/java/com/step902020/capstone/UserServlet.java
meltsufin: Get, modify, save should work.

capstone/src/main/resources/static/savedevents.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta>
+    <title>My Account</title>
+    <link rel="stylesheet" href="style.css">
+    <script src="script.js"></script>
+</head>
+<body onload="getUser();">
+    <nav>
```
meltsufin, line 10: The whole nav should become an include eventually so that you don't have to duplicate it on every html page.
It will also require you to switch to server-side templating like JSP or Thymeleaf. Can you file it as a separate issue to take care of in the future?

capstone/src/main/resources/static/script.js
```diff
@@ -0,0 +1,76 @@
+
+/** checks whether the user is authenticated and adjust elements 
+according to whether the user is logged in or logged out */
+function checkAuth(){
+  // send request for information on login status
+  fetch('_gcp_iap/identity').then(response => response.json()).then((data) => {
+    const userEmail = document.getElementById('email-form');
+    // default value set for testing on dev server
+    const email = "accounts.google.com:jennysheng@google.com";
```
meltsufin, line 9: I understand you want to be able to run locally. So, is it possible to detect that data is empty, and only then hard-code the email?

capstone/pom.xml
jsheng112: I'm not sure what this means. Am I using the wrong dependency?

capstone/src/main/java/com/step902020/capstone/UserServlet.java
jsheng112: I could not find the documentation for modify, but it seems that once you have the entity id and that id remains the same then "save" does an update instead

jsheng112:
I did a pretty significant renaming of documents and classes (because I was getting confused so I thought it would be best to fix the issue sooner than later). As a result, some of the documents completely highlighted in green are not actually new code, but old files with names changed.

Naming patterns: "individual" refers to individual users, "organization" refers to organizations, "user" refers to organizations and individuals (this appears sometimes because some functions are used for both types)

capstone/src/main/java/com/step902020/capstone/Organization.java
```diff
@@ -0,0 +1,83 @@
...
+  
+  public Organization(Long timestamp, String name, String email, String university, String userType, String description, String image) {
+    this.timestamp = timestamp;
+    this.name = name;
+    this.email = email;
+    this.university = university;
+    this.userType = userType;
+    this.description = description;
+    this.image = image;
+  }
```
lbourret, line 38: if a user wants to just update only one detail, is there a way to prefill the form so they dont have to rewrite all the details?

capstone/src/main/java/com/step902020/capstone/Organization.java
```diff
@@ -0,0 +1,83 @@
...
+  String userType;
+  String description;
+  String image;
+  
+  public Organization() {
+
+  }
+  
+  public Organization(Long timestamp, String name, String email, String university, String userType, String description, String image) {
+    this.timestamp = timestamp;
```
lbourret, line 31: also just wondering, is there a reason to collect timestamps?

capstone/src/test/java/com/step902020/capstone/UserDatastoreTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.boot.test.web.client.TestRestTemplate;
+import org.springframework.boot.web.server.LocalServerPort;
+import java.util.Random;
+import org.junit.Before;
+import org.junit.After;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = 
+		CapstoneApplication.class, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+public class UserDatastoreTest {
```
lbourret, line 34: way to add tests!

capstone/pom.xml
meltsufin: You just need to add 
`<scope>test</scope>` so that this dependency doesn't get bundled into the jar that you deploy, since it's only used during testing.

capstone/src/test/java/com/step902020/capstone/UserDatastoreTest.java
```diff
@@ -0,0 +1,93 @@
...
+    final String baseUrl = "http://localhost:"+ port + "/get-organization?email=" + expectedEmail;
+    URI uri = new URI(baseUrl);
+ 
+    Organization[] result = restTemplate.getForObject(uri, Organization[].class);
+
+    Assert.assertEquals("Wrong user returned", expectedEmail, result[0].getEmail());
+  }
+
+  @Before
+  public void setUp() {
```
meltsufin, line 80: Generally, `setUp` and `tearDown` are at the top of the class before all of the `@Test` methods come, to aid readability.

capstone/src/main/java/com/step902020/capstone/Organization.java
jsheng112: Its not really relevant to us right now, but timestamps usually help to visualize growth (like how many users registered and started using the app in a month). I figured it would be a pain to add it later if we ever need it.

capstone/src/main/java/com/step902020/capstone/UserController.java
```diff
@@ -0,0 +1,81 @@
...
+  public RedirectView saveIndividual(
+      @RequestParam("id") String id,
+      @RequestParam("firstname") String firstname,
+      @RequestParam("lastname") String lastname, 
+      @RequestParam("email") String email, 
+      @RequestParam("user-type") String userType,
+      @RequestParam("university") String university) throws IOException {
+
+    // each email can only exist as either an individual or a user, not both
+    this.organizationRepository.deleteByEmail(email);
```
elefeint, line 46: Non-actionable for now, since this is test data: in real life, this would enable a disgruntled student deleting an organization's profile by creating a personal profile with the same e-mail address.

capstone/src/main/java/com/step902020/capstone/UserController.java
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+      current = new Individual(Long.parseLong(id), System.currentTimeMillis(), firstname, lastname, email, university, userType, "");
+    }
+    this.individualRepository.save(current);
+    return new RedirectView("profile.html", true);
+  }
+
+  @PostMapping("save-organization")
+  public RedirectView saveOrganization(
+      @RequestParam("id") String id,
```
elefeint, line 61: You can use `@RequestParam("id") Optional<String> id` instead of comparing to an empty string.

capstone/src/main/java/com/step902020/capstone/UserController.java
jsheng112: Actually this is something I was struggling with. Since we are storing individuals and users separately I need to check for duplicate accounts in the other user type. I also need to fetch from both entities. I was wondering if there's some way around it? Like query both entities at the same time through a common property?

meltsufin: It might be easier to keep organizations and individuals more separate. For instance, you can have a separate controller for organizations, and even separate forms/pages for management.

elefeint: They are different entity types, so you can't query them in one go. But querying both and failing with a meaningful message if you find an existing entity of a different type is better than silently deleting the conflicting entry.

PR #9, opened 2020-06-26
Basic Event Feed Page

lbourret:
Minimum outline & haven't implemented filter
Set up temp css and event class to prevent merge conflicts with Jenny (CSS) and Olivia (Event)


meltsufin:
@lbourret The official guidance is that both of your pod-mates should review each PR before we (the hosts) do the final review. It may slow down the turnaround, but we should give it a shot.

capstone/src/main/resources/static/index.html
```diff
@@ -0,0 +1,70 @@
...
+    <meta charset="UTF-8">
+    <title>Events</title>
+    <link rel="stylesheet" href="style1.css">
+    <script src="script.js"></script>
+  </head>
+
+  <body>
+    <h1>Events</h1>
+    <div id="filters">
+      <p>Filters</p>
```
jsheng112, line 13: The format of the filters is a little confusing. Maybe place them in a column or put some spaces between them, but feel free to address this in a later PR since style changes is not in the scope of this PR.

capstone/src/main/resources/static/script.js
```diff
@@ -0,0 +1,65 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Retrieves events from server
+ */
+function getEvents() {
+  fetch('/list-events').then(response => response.json()).then((events) => {
```
jsheng112, line 19: For these mappings maybe it is a good idea to use relative paths instead of absolute paths (remove the "/" in the front) but I'm not sure if spring-boot supports relative paths.

capstone/src/main/resources/static/script.js
```diff
@@ -0,0 +1,65 @@
...
+  // Name 
+  const nameElement = document.createElement('p');
+  nameElement.innerText = event.text;
+
+  /*
+  Time
+  Location
+  Organization
+  Description
+  */
```
jsheng112, line 44: Good idea! These comments show very clearly what fields are going to be displayed in the future. Also helps with planning and understanding the code.

capstone/src/main/resources/static/index.html
```diff
@@ -0,0 +1,70 @@
...
+
+          <label for="vistor">Visitors Allowed</label>
+          <input type="checkbox" id="vistor" name="vistor" value="vistor">
+        </div>
+
+        <label for="sort">SORT BY</label>
+        <select name="sort" id="sort" class="button smallDefaultButton">
+            <option value="descending">NEWEST</option>
+            <option value="ascending">OLDEST</option>
+        </select>
```
oasemota99, line 60: I think the Sort By function is great! I think adding other ways to sort will definitely be a plus later on, but I don't know how necessary sorting by oldest will be? My assumption is that a user will always want the latest events, but the functionality never hurts

capstone/src/main/java/com/step902020/capstone/EventTemp.java
```diff
@@ -0,0 +1,26 @@
+package com.step902020.capstone;
+
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Entity;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Field;
+import org.springframework.data.annotation.Id;
+
+@Entity(name = "eventTemp")
+public class EventTemp {
```
oasemota99, line 8: Haha I love the bare-bones functionality here, I'll be sure to get you something to actually work with soon!

capstone/src/main/resources/static/script.js
elefeint: When specified from Javascript, the relative vs absolute path will affect which back-end URL will be requested. Let's say, your domain is _collegeevents.com_, and your application sits in the very root of the domain. In this case, the distinction between an absolute and a relative URL won't matter -- either way, if you refer to "event/list" or "/event/list", the fetch function will contact _collegeevents.com/event/list_.

However, let's say you have multiple applications running on the same domain, and you want the event finding/rating application to live at _collegeevents.com/eventsapp_. In this case, "event/list" will go to _collegeevents.com/eventsapp/event/list_, but "/event/list" will go to the root of the domain -- _collegeevents.com/event/list_ where it won't find any of the controllers (since the server-side app listens to URLs under the `event` grouping.

For the purposes of the current app, you can do it either way, just be consistent. I wanted you to be aware of the distinction because you'll some day run into troubleshooting why your events go to the root of the domain.

capstone/src/main/resources/static/index.html
lbourret: oh oops that was leftover from copying the portfolio code!

elefeint: Code review; working as intended. You've all done a great job reviewing!

capstone/src/main/resources/static/index.html
```diff
@@ -0,0 +1,70 @@
...
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Events</title>
+    <link rel="stylesheet" href="style1.css">
+    <script src="script.js"></script>
+  </head>
+
+  <body>
+    <h1>Events</h1>
```
meltsufin, line 11: I like how bare bones this is to start with. You can always enhance it and add more css and headers etc. later!

PR #31, opened 2020-06-30
Save events and organizations

jsheng112:
I combined add, delete, and display saved events into one PR. UserController is separated into IndividualController and OrganizationController to make the two types of users more distinct. 
Now, when a new user logs in to the website there are two different forms to toggle between. When an old user logs in the page will display a form that fits their user type.


capstone/src/main/resources/static/profile.html
```diff
@@ -44,18 +46,16 @@ <h1>My Profile</h1>
...
     </div>
     <div class="profile-form general-container" id="user">
         <form action="save-individual" method="POST" id="user-form">
-            <label>Are you an individual or a club/organization?</label>
-            <select  id="user-type-toggle" name="user-type" onchange="toggleForm('user-type-toggle');">
-                <option value="individual">Individual</option>
-                <option value="organization">Organization</option>
-            </select>
+            <div id="user-select">
+                <label>Are you an individual or a club/organization?</label>
```
lbourret, line 53: minor but I think you could just write "Individual or Club" just like you would only write "First Name"

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -0,0 +1,85 @@
...
+
+    Individual current = null;
+    List<Individual> userList = this.individualRepository.findByEmail(email);
+    
+    // either edit the existing user or create a new one
+    if (userList.size() > 0) {
+      current = userList.get(0);
+      current.editFirstName(firstname);
+      current.editLastName(lastname);
+      current.editUniversity(university);
```
lbourret, line 45: Right now, it seems like if a user wanted to changed one details, they would have to resubmit all details or else it will all be overwritten with "" ie. if I changed my first name, but didn't fill out the whole form, my last name and university values would be empty

capstone/src/main/java/com/step902020/capstone/Individual.java
```diff
@@ -41,17 +44,10 @@ public Individual(Long timestamp, String firstName, String lastName, String emai
...
-    this.firstName = firstName;
-    this.lastName = lastName;
-    this.email = email;
-    this.university = university;
-    this.userType = userType;
-    this.image = image;
+    savedEvents = new HashSet<String>();
+    savedEvents.add("hello 1");
+    savedEvents.add("hello 2");
+    savedEvents.add("hello 3");
```
lbourret, line 50: Do you still need the hard coded saved events?

capstone/src/main/java/com/step902020/capstone/Individual.java
```diff
@@ -85,4 +81,32 @@ public String getUserType() {
...
+
+  public void editFirstName(String firstName) {
+    this.firstName = firstName;
+  }
+
+  public void editLastName(String lastName) {
+    this.lastName = lastName;
+  }
+
+  public void editUniversity(String university) {
```
lbourret, line 97: Should users be able to change their universities?

capstone/src/main/java/com/step902020/capstone/Organization.java
```diff
@@ -79,5 +67,21 @@ public String getDescription() {
   public String getImage() {
     return image;
   }
+
+  public void editName(String name) {
+    this.name = name;
+  }
+
+  public void editUniversity(String university) {
```
lbourret, line 75: Should orgs be able to change their affiliated college?

capstone/src/main/java/com/step902020/capstone/Individual.java
jsheng112: Yes for now. So far it doesn't seem like the "Event" entity is complete yet.

capstone/src/main/java/com/step902020/capstone/Individual.java
jsheng112: You're right it's probably not a good idea.

capstone/src/main/java/com/step902020/capstone/Individual.java
```diff
@@ -85,4 +81,32 @@ public String getUserType() {
   public String getImage() {
     return image;
   }
+
+  public Set<String> getSavedEvents() {
+    return savedEvents;
+  }
+
+  public void editFirstName(String firstName) {
```
elefeint, line 89: General java bean naming convention is `setFirstName()`

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -0,0 +1,85 @@
...
+  @PostMapping("save-individual")
+  public RedirectView saveIndividual(
+      @RequestParam("firstname") String firstname,
+      @RequestParam("lastname") String lastname, 
+      @RequestParam("email") String email, 
+      @RequestParam("user-type") String userType,
+      @RequestParam("university") String university) throws IOException {
+
+    Individual current = null;
+    List<Individual> userList = this.individualRepository.findByEmail(email);
```
elefeint, line 38: I missed this in the previous PR. If `findByEmail()` is always expected to return at most one result, you can return `Optional<Individual>` instead of a list. It will simplify the logic for retrieving the current user in a few places.
https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repository-query-return-types

capstone/src/main/resources/static/script.js
```diff
@@ -83,40 +82,74 @@ function checkAuth(){
...
-function getUser() {
+/* get and store the user type (individual or organization). With this we determine whether
+this user exists in our database or not */
+function getUserType() {
     var email = checkAuth();
 
     /* Since there is no way to know beforehand whether the user is an organization 
     or an individual, we have to do two fetches to check the organization entities and
     the user entities */
     fetch('get-organization?email=' + email).then(response => response.json()).then((data) => {
```
elefeint, line 134: Out of scope for this PR: consider introducing a new server-side endpoint `get-user-type`; then you won't need to sequential calls.

capstone/src/main/resources/static/script.js
```diff
@@ -83,40 +82,74 @@ function checkAuth(){
...
+  var userType = sessionStorage.getItem("user-type");
+
+  // if there is no userType stored in this session, that means this is a new user
+  if (userType == null) {
+    getUserType();
+    displayForm("individual", true);
+  } else {
+    if (userType == "organization") {
+      fetch('get-organization?email=' + email).then(response => response.json()).then((data) => {
+        createProfile(data, true);
```
elefeint, line 107: The next 3 lines can probably be a helper method; they repeat in both branches.

capstone/src/main/resources/static/script.js
```diff
@@ -185,32 +219,68 @@ function createOrgProfile(data) {
...
+function toggleForm(formUserType) {
+  var userType = document.getElementById(formUserType).value;
+  displayForm(userType, true);
+}
+
+// helper function for displaying forms
+// a new user will be able to toggle, but a returning user will not
+function displayForm(userType, displayBoth) {
   if (userType == "individual") {
     document.getElementById("user").style.display = "block";
```
elefeint, line 232: There is likely an opportunity here to have a helper method `updateUserType(isOrganization)`. Whenever you find code blocks that look almost identical except for some values being set, consider a helper method that takes a parameter differentiating those values.

capstone/src/main/java/com/step902020/capstone/Individual.java
lbourret: Oh sorry I meant that since you're storing a set<String>, you can populate the list with the String values from the save-event input box so you wouldn't need to hardcode strings. but since it'll temp anyways it doesn't matter which way you choose!

capstone/src/main/resources/static/script.js
jsheng112: How would this look with the two separate controllers that I have? If I have this in one controller the other user type can't be searched. Or would this reside in a separate controller?

elefeint: Separate controller, I think. On the backend it will still have to check both people and organizations, but at least your front-end will have a clean abstraction.

// non-actionable aside
In database design, this kind of problem where some fields are shared and some are not could be solved with 3 tables: "user", "person-user" and "organization-user". The "user" table would have a column for type, and the other two tables would contain fields not relevant to the other type.
As a completely different solution, you could solve this on the front-end by having two completely distinct screens for people and organizations. 
So, there are many ways to approach the same usecase.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
jsheng112: If I do not find a user and it returns an optional.empty() how do I detect this on the front end when it is turned into a json? I couldn't find documentation for this.

capstone/src/main/resources/static/script.js
jsheng112: Filed this as an issue to be addressed in a later PR.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
elefeint: Ah, great question! It might require some experimentation. I think I would return `optionalResult.orElse(null)`, hoping that null will get serialized as an empty Json object. So the REST controller return type would just be `Individual`.
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElse-T-

I had to google around for how Jackson (which Spring Web uses to serialize) deals with Optional types: https://www.baeldung.com/jackson-optional

jsheng112: Sorry Elena, I think you might have seen the previous version of my comment. I was trying to see what value optional.empty() would return on the frontend but I edited the comment after I realized the Java version is incorrect on cloud shell. I don't know if I should go ahead and update Java.

capstone/src/main/resources/static/profile.html
```diff
@@ -44,6 +44,7 @@ <h1>My Profile</h1>
                 <div id="user-type"></div>
                 <div id="university"></div>
                 <div id="description"></div>
+                <button class="button submit" onclick="revealForm()">Edit/Submit Your Profile</input>
```
lbourret, line 47: i like the toggle edit form!

capstone/src/main/resources/static/script.js
```diff
@@ -100,29 +100,31 @@ function getUser() {
...
-        document.getElementById("no-profile").style.display = "none";
         sessionStorage.setItem("user-type", data[0].userType);
       });  
     }
-    // display profile
+    displayMain(true);
+  }
+}
+
+function displayMain(display) {
```
lbourret, line 119: good idea for helper function

capstone/src/main/resources/static/script.js
```diff
@@ -158,7 +157,6 @@ function getUserType() {
 function createProfile(data, isOrganization) {
```
lbourret, line 157: Might be more readable if you pass the user, not the data so you don't have to write data[0] each time

capstone/src/main/resources/static/style.css
```diff
@@ -141,13 +140,11 @@ body {
...
 
 /* for changing color on hover in the nav  */
-li a:hover {
-  background-color: #1B70D3;
-}
- 
-/* for changing color when clicked in the nav */
-.active {
-  background-color: #64BBD7;
+li a:hover, .active a {
```
lbourret, line 143: I like the inverse!

capstone/src/main/resources/static/script.js
jsheng112: Good point probably would make the code clearer too.

capstone/src/test/java/com/step902020/capstone/UserDatastoreTest.java
```diff
@@ -27,6 +27,13 @@
 import java.util.Random;
 import org.junit.Before;
 import org.junit.After;
+import java.net.URL;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpEntity;
+import org.json.JSONObject;
```
meltsufin, line 33: This is an unused import, among several others in this file.
IntelliJ clearly shows this to me and can easily auto-fix it, but I know you're using Cloud Shell which makes it difficult.

capstone/pom.xml
```diff
@@ -64,6 +64,11 @@
       <version>3.3.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
```
meltsufin, line 69: I don't believe you're using this anywhere.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -0,0 +1,114 @@
...
+  private IndividualRepository individualRepository;
+
+  @GetMapping("get-individual")
+  public List<Individual> getIndividual(@RequestParam("email") String email) {
+    return this.individualRepository.findByEmail(email);
+  }
+
+
+  @PostMapping("save-individual")
+  public RedirectView saveIndividual(
```
meltsufin, line 30: In terms of naming of methods and fields, I find the use of the word "save" to be a bit ambiguous because it could mean either "create" or "update".
For this specific method, `registerIndividual` would probably be more clear.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -0,0 +1,114 @@
...
+    else {
+        current = new Individual(System.currentTimeMillis(), firstname, lastname, email, university, userType, "");
+    }
+    this.individualRepository.save(current);
+    return new RedirectView("profile.html", true);
+  }
+
+
+  @PostMapping("add-saved-event")
+  public RedirectView addSavedEvent(
```
meltsufin, line 55: It's a good practice to add javadocs to all public methods.
It helps explain what each parameter is for and what the method does.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -0,0 +1,114 @@
...
+      @RequestParam("event-name") String eventName,
```
meltsufin, line 56: Will this eventually be replaced with an event ID?

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -0,0 +1,114 @@
...
+      @RequestParam("email") String email) throws IOException {
```
meltsufin, line 57: Security question: If the email is of the user for whom the event is being saved, what prevents anyone from adding an event for any user?

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+    this.individualRepository.save(current);
+    return new RedirectView("savedevents.html", true);
+  }
+  
+
+  @PostMapping("delete-saved-event")
+  public RedirectView deleteSavedEvent(
+      @RequestParam("event-name") String eventName,
+      @RequestParam("email") String email) throws IOException {
```
meltsufin, line 73: Why not use user ID (`datastoreId`) in all of these methods instead of email?
If you ID, you can just use "findById()".

capstone/src/main/java/com/step902020/capstone/OrganizationController.java
```diff
@@ -0,0 +1,66 @@
...
+
+/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+@RestController
+public class OrganizationController {
+
+  @Autowired
+  private OrganizationRepository organizationRepository;
+
+  @GetMapping("get-organization")
+  public List<Organization> getOrganization(@RequestParam("email") String email) {
```
meltsufin, line 24: Why would there ever be more than one organization with a specific email?
Also, I would probably prefer using `datastoreID` over email because an email of an organization can potentially be updated, but `datastoreID` is a better permanent primary key.

capstone/src/test/java/com/step902020/capstone/UserDatastoreTest.java
```diff
@@ -89,4 +96,122 @@ public void testGetOrganization() throws URISyntaxException {
...
+    Set<String> expectedSet = new HashSet<>();
+    expectedSet.add("test");
+ 
+    Individual[] result = restTemplate.getForObject(uri, Individual[].class);
+
+    Assert.assertEquals("Insert event error", expectedSet, result[0].getSavedEvents());
+  }
+
+  @Test
+  public void testDeleteEvent() throws URISyntaxException {
```
meltsufin, line 131: It might not be immediately obvious, but there is an opportunity in these tests to make them less repetitive with helper functions. Notice the lines that are identical or almost identical across the test methods.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
jsheng112: Yeah I agree, but I guess the ambiguity is there because this method does either an update or a create depending on whether the individual exists already.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
jsheng112: At all point in time, the user is not able to change emails. Before these post requests, we do a "checkAuth()" to get the email of the user through IAP.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
jsheng112: Compared to the id, the user email is readily available because before any fetch requests I have to do a checkAuth() to make sure the user is properly logged in.

capstone/src/main/java/com/step902020/capstone/OrganizationController.java
jsheng112: There usually is just one entity returned. But sometimes none are found. Right now I handle this by passing the whole list and then checking for the list size on the front end. Is it better to do it this way or for me to pass just the Organization object, which may or may not be null.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
meltsufin: A malicious user would still be able to run `curl` command with a modified email address.
Checking the email in JavaScript is not sufficient for security.
Instead, you should extract the email address from the HTTP header that IAP sets.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
jsheng112: Should I also add comments to the getters and setters in Individual.java and Organization.java?

meltsufin: You can skip getters and setters. I would generally be pragmatic and focus on methods that benefit from an explanation.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
jsheng112: Yes I definitely need to change it to an event Id.

capstone/src/test/java/com/step902020/capstone/UserDatastoreTest.java
jsheng112: I actually have some issues with these tests. I changed all the mappings to take in a RequestHeader param from the iap email header. But that causes these tests to fail locally. I have commented them out for now and I will create an issue to address these repetitive lines.

jsheng112:
I changed all the IAP headers on the backend, therefore the application does not work locally. I have deployed this branch into the step90-2020 project.

PR #46, opened 2020-07-07
Base event creation

oasemota99:
So, this PR is a little messy, but I wanted to get something out so you all could try to test out your features. Any code that is commented out will be implemented later, but I'm going to need a bit more time to make sure it functions properly.
With this PR, you can create an event using the following parameters:
- Organization Name
- Event Title
- Event Time and Date
- Event Description
- Event Longitude and Latitude

Side Note: Each event you submit can be seen in datastore, but it doesn't properly redirect you back to the event.html page (working on a fix for that as well)


oasemota99:
I commented out the merge conflicts instead of just erasing them, @lbourret  @jsheng112  in case you all wanted to use your temp code

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.*;
+
+@Entity(name = "event")
+public class Event{
+
+  @Id
+  Long datastoreID;
+
+  @Field(name="organizationName")
+  String organizationName;
```
meltsufin, line 16: You can use an `Organization` reference here with `@Reference`.
See: https://cloud.spring.io/spring-cloud-gcp/reference/html/#key-reference-relationships

capstone/src/main/java/com/step902020/capstone/EventRepository.java
```diff
@@ -1,11 +1,14 @@
...
-/**
- * 
- */
-public interface EventRepository extends DatastoreRepository<EventTemp, Long> {
-    public EventTemp findById(long id);
-}
\ No newline at end of file
+public interface EventRepository extends DatastoreRepository<Event, Long> {
+
+  public List<Event> findByDatastoreID(Long datastoreID);
```
meltsufin, line 9: I believe this is equivalent to `findById` which already exists in the `DatastoreRepository`.

elefeint: Ah! It likely is. I've misled @oasemota99 by changing the custom interface method from `findByID` to `findByDatastoreID`, but removing the method entirely and using the built-in one is likely a better solution!

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -0,0 +1,95 @@
...
+  // String university;
+
+  @Field(name="eventLatitude")
+  double eventLatitude;
+
+  @Field(name="eventLongitude")
+  double eventLongitude;
+
+  // @Field(name="eventFilters")
+  // List<String> eventFilters;
```
elefeint, line 37: Would those be filters or "categories"?

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -0,0 +1,95 @@
...
+    return datastoreID;
+  }
+
+  public String getEventTitle() {
+    return eventTitle;
+  }
+  public String getOrganizationName() {
+    return organizationName;
+  }
+  public String geteventDateTime() {
```
elefeint, line 76: Spring framework's Json conversion uses getters to go from an object to a JSON text, relying on java bean naming conventions. I am not sure if it can handle uncapitalized `event` (it might?).
But it's best to have strict camel case for getters/setters, so `getEventDateTime`.

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -0,0 +1,95 @@
...
+  public String getDescription() {
+    return eventDescription;
+  }
+  // public String getuniversity() {
+  //   return university;
+  // }
+  public double getEventLatitude() {
+    return eventLatitude;
+  }
+  public double geteventLongitude() {
```
elefeint, line 88: Same here: `getEventLongitude`.

capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -24,27 +27,53 @@
...
+    @RequestParam("organizationName") String organizationName,
+    @RequestParam("eventTitle") String eventTitle,
+    @RequestParam("eventDateTime") String eventDateTime,
+    @RequestParam("eventDescription") String eventDescription,
+    @RequestParam("eventLatitude") String eventLatitude,
+    @RequestParam("eventLongitude") String eventLongitude
+    // @RequestParam("eventFilters") List<String> eventFilters
+    ) throws IOException {
+      
+      Event newEvent = new Event(organizationName, eventTitle, eventDateTime, eventDescription, Double.parseDouble(eventLatitude), Double.parseDouble(eventLongitude));
```
elefeint, line 71: In the future, you'll need to do some data validation on the coordinates -- what if the front-end passes empty strings for those?

capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -24,27 +27,53 @@
...
+
+      this.eventRepository.save(newEvent);
+      return new RedirectView("event.html", true);
```
elefeint, line 74: A possible enhancement (you can create a follow-up issue; it's not critical) is to redirect to that specific event.

capstone/src/main/resources/static/event.js
```diff
@@ -0,0 +1,51 @@
...
+
+  // const organizationID = document.createElement('p');
+  // organizationID.innerText = event.organizationID;
+  // listEventElement.appendChild(organizationID);
+
+  const organizationName = document.createElement('p');
+  organizationName.innerText = event.organizationName;
+  listEventElement.appendChild(organizationName);
+
+  const eventTitle = document.createElement('p');
```
elefeint, line 21: These 3 lines repeat a lot. There is an opportunity here to create a helper method!

capstone/src/main/java/com/step902020/capstone/Event.java
oasemota99: Categories is probably a better description, but I used 'filters' for continuity's sake

capstone/src/main/resources/static/event.html
```diff
@@ -40,17 +37,17 @@ <h1>Create an Event </h1>
         <br><br>
 
         <label for="eventLocation"> Event Location: </label> <br>
-        <input type="number" id="eventLat" name="eventLat" placeholder="Event Latitude"> <br>
-        <input type="number" id="eventLng" name="eventLng" placeholder="Event Longitude">
+        <input type="number" step="0.000001" id="eventLatitude" name="eventLatitude" placeholder="Event Latitude"> <br>
+        <input type="number" step="0.000001" id="eventLongitude" name="eventLongitude" placeholder="Event Longitude">
```
jsheng112, line 41: Works great! This is definitely out of the scope of this PR but we should consider adding like a mini map of some sort that takes a marker instead of having the user type in coordinates.

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -0,0 +1,95 @@
...
+  //   this.organizationID = organizationID;
+  //   this.organizationName = organizationName;
+  //   this.eventTitle = eventTitle;
+  //   this.eventDateTime = eventDateTime;
+  //   this.eventDescription = eventDescription;
+  //   this.university = university;
+  //   this.eventLatitude = eventLatitude;
+  //   this.eventLongitude = eventLongitude;
+  //   this.eventFilters = eventFilters;
+  // }
```
jsheng112: I would start uncommenting these lines of code. Don't worry about conflicts with Lian or I. We have to shift over to the actual event class eventually so better to do it sooner than later.

capstone/src/main/resources/static/event.html
jsheng112, line 41: Works great! This is definitely out of the scope of this PR but we should consider adding like a mini map of some sort that takes a marker instead of having the user type in coordinates.

capstone/src/main/java/com/step902020/capstone/EventController.java
oasemota99: noted!

capstone/src/main/java/com/step902020/capstone/EventController.java
oasemota99: I will look into it!

capstone/src/main/resources/static/event.html
oasemota99: Yeah, I will definitely look into it!

PR #58, opened 2020-07-08
adds a basic user calendar

jsheng112:
This whole PR focuses on layout and style changes. 
The calendar is non functional and only serves as a starting point for laying out the calendar.


capstone/src/main/resources/static/style.css
```diff
@@ -62,8 +70,21 @@ img {
   padding: 15px;
 }
  
+.profile-head {
```
lbourret, line 73: profile looks SO nice!

capstone/src/main/resources/static/script.js
```diff
@@ -416,3 +403,20 @@ function closeForm() {
 	document.getElementById("organization").style.display = "none";
 }
 
+/* function to generate divs for the calendar */
+function createCalendar() {
+  const calendar = document.getElementById("calendar");
+  for (var i = 0; i < 14; i++) {
+    var nextDay = new Date();
+    var today = new Date();
+    nextDay.setDate(today.getDate() + i);
```
lbourret, line 412: date has some different formatting that might be helpful to cut down on the amount of info given with the time

capstone/src/main/resources/static/script.js
```diff
@@ -416,3 +403,20 @@ function closeForm() {
 	document.getElementById("organization").style.display = "none";
 }
 
+/* function to generate divs for the calendar */
+function createCalendar() {
+  const calendar = document.getElementById("calendar");
+  for (var i = 0; i < 14; i++) {
```
lbourret, line 409: Not something we need to address in this pr but what do you think of changing the design from having for each week instead of each day? Honestly have NO idea how that'd turn out but right now having a block for each day feels a bit cluttered.

jsheng112: Yeah, I actually tried formatting it into a seven column (so like by week instead of by day), and it also seems equally cluttered to be honest because then the events would be formatted vertically one after the other in a column. I'm not sure what to do.

capstone/src/main/resources/static/usercalendar.html
```diff
@@ -0,0 +1,73 @@
...
+                <p> event 5</p>
+            </div>
+            <div class="event general-container col">
+                <p> event 1</p>
+            </div>
+            <div class="event general-container col">
+                <p> event 2</p>
+            </div>
+            <div class="event general-container col">
+                <p> event 3</p>
```
oasemota99, line 67: Typo here? I'm assuming you meant event 6, 7, and 8?

jsheng112: Not really a typo, the events are there for now to demonstrate the layout. I basically copied and pasted the same three events over.

PR #63, opened 2020-07-09
Organization profile

### LINKED ISSUE #41: Create manage event page for organizations

### LINKED ISSUE #51: user references instead of datastoreId

jsheng112:
- The website toggles between displaying two different nav bars for the two types of users.
- Organizations can now access their events through the "manage my events" page, and edit
events through the same form as creating the events.
- I moved the js functions pertaining to user accounts from script.js into user.js. (getOrganizationEvents() is the only new method)
- https://jenny-dot-step90-2020.ue.r.appspot.com/ has this current version deployed for testing.
- testing: to test, you would need to close existing tabs of the website (because I use session storage), go to datastore console and delete any instance of your email (because some properties changed). Then sign into the website and create an organization account.
- fixes #41 and fixes #51.


capstone/pom.xml
```diff
@@ -84,7 +84,14 @@
...
         <configuration>
           <deploy.projectId>step90-2020</deploy.projectId>
-          <deploy.version>1</deploy.version>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>11</source>
```
meltsufin, line 93: I think the build is failing because you're using Java 11 and CI build is configured to use Java 8.
Why did you change to Java 11?

meltsufin: See: https://github.com/googleinterns/step90-2020/blob/master/.github/workflows/maven.yml

meltsufin: I updated the CI in #65, but you probably need to sync to master to get the effect.

jsheng112: Thanks for finding the problem for me. I was struggling to identify where the issue was because it was building and running fine locally. This is so odd because I never changed it to Java 11.

meltsufin: I wonder how `maven-compiler-plugin` got in the pom then. Maybe IntelliJ inserted it?
In any case, in makes sense to be on 11 because we're deploying to App Engine that runs Java 11.

lbourret:
fixes #79

capstone/src/main/resources/static/user.js
```diff
@@ -0,0 +1,375 @@
...
+  nameContainer.appendChild(pElementName);
+
+  // hide fields that pertain to organizations only
+  document.getElementById("description").style.display = "none";
+
+  if (fillForm) {
+    // prefill form
+    document.getElementById("ind-firstname").value = data.firstName;
+    document.getElementById("ind-lastname").value = data.lastName;
+    document.getElementById("university-form-display").innerText = data.university;
```
lbourret, line 108: oh shoot in my pr, I hide this element bc it was creating white space in the form. I had it "revealed" in this function but in script.js: `document.getElementById("university-form-display").display = 'block';`

capstone/src/main/resources/static/savedorganizations.html
```diff
@@ -4,7 +4,7 @@
     <meta>
     <title>My Account</title>
     <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
+    <script src="user.js"></script>
```
lbourret, line 7: nice, these files were getting huge

capstone/src/main/resources/static/manageevents.html
```diff
@@ -0,0 +1,52 @@
...
+        <div class="container">
+            <a href="images/js7.jpg"><img class="image profile" src="images/js7.jpg"/></a>
+        </div>
+        <div id="email"></div>
+        <div id="university"></div>
+        <div id="description"></div>
+        <div id="no-profile">
+            <p>You have not built your profile yet, please submit a profile </p>
+        </div>
+        <nav class="nav-bar-inverted" id="individual-nav">
```
lbourret, line 31: Should this be org-nav? Since manageevents.html seems like a org profile page?

capstone/src/main/resources/static/manageevents.html
```diff
@@ -0,0 +1,52 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta>
+    <title>My Account</title>
+    <link rel="stylesheet" href="style.css">
+    <script src="user.js"></script>
+</head>
+<body onload="getUser(false);getOrganizationEvents();">
```
lbourret, line 9: I ran into TypeError for forEach with this function

capstone/src/main/resources/static/user.js
```diff
@@ -0,0 +1,375 @@
...
+/* function to get all events hosted by the current organization */
+function getOrganizationEvents() {
+  var userType = sessionStorage.getItem("user-type");
+  if (userType == null) {
+    getUserType();
+  }
+  if (userType == "organization") {
+    fetch('get-' + userType).then(response => response.json()).then((data) => {
+      const eventDiv = document.getElementById("created-events");
+      data[0].events.forEach((event) => eventDiv.appendChild(createSavedEventElement(event), false));
```
lbourret, line 332: Ok I did a super quick poke around and I think the error is that there the events list is null for most orgs

capstone/src/main/resources/static/user.js
jsheng112: Ohh yeah, it basically is a toggle between this university p element and the university input element depending on whether the user had entered a university before.

capstone/src/main/resources/static/manageevents.html
jsheng112: No it should be individual nav. Each of the pages has two nav bars that gets hidden or displayed based on user type. The second nav is right below this one.

capstone/src/main/resources/static/user.js
jsheng112: Yeah I just realized that most of the test data for organizations are out of date and do not have the required fields. I will try to add some on Monday.

jsheng112: Sorry I don't want to leave you hanging with the error. I'm not sure if you used an old account instead of creating a new one when testing, but I think you got the error because some of the fields for organizations changed types. The error would resolve itself if you delete the current account and create a new one.

lbourret: works!

elefeint:
This PR had such a great description!

PR #73, opened 2020-07-10
Basicfilters

oasemota99:
Small PR, just wanted to get a couple of working event filters in


capstone/src/main/java/com/step902020/capstone/EventRepository.java
```diff
@@ -1,7 +1,7 @@
 package com.step902020.capstone;
 
 import org.springframework.cloud.gcp.data.datastore.repository.DatastoreRepository;
-
+import org.springframework.core.convert.converter.Converter;
```
jsheng112, line 4: I don't think this is used anywhere.

capstone/src/main/resources/static/event.js
```diff
@@ -17,6 +17,8 @@ function createEvent(event) {
   createEventAttribute(event.eventDescription, listEventElement);
   createEventAttribute(event.eventLatitude, listEventElement);
   createEventAttribute(event.eventLongitude, listEventElement);
+  createEventAttribute(event.foodAvaliable, listEventElement);
+  createEventAttribute(event.requiredFee, listEventElement);
```
jsheng112, line 21: Very concise and useful helper function!

lbourret: ^

capstone/src/main/resources/static/event.html
```diff
@@ -49,28 +43,22 @@ <h1>Create an Event </h1>
         <label for="eventDescription"> Event Description: </label> <br>
```
lbourret, line 43: Sorry I couldn't comment on a line that you didn't change, but on line 38 you don't have a counterpart for "for="eventlocation"" (would have ever noticed it but intellij caught it)

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -30,17 +30,27 @@
 
   @Field(name="eventLongitude")
   double eventLongitude;
+
+  @Field(name="foodAvaliable")
+  boolean foodAvaliable;
+
+  @Field(name="requiredFee")
+  boolean requiredFee;
```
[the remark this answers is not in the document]
elefeint: If you put them in a list, they'll be harder to query for individually in datastore.
Here are the kinds of filters supported: https://cloud.google.com/datastore/docs/concepts/queries

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -30,17 +30,27 @@
...
   
   @Reference
   List<Review> reviews;
+ 
 
-  public Event(Organization organization, String eventTitle,String eventDateTime, String eventDescription, double eventLatitude, double eventLongitude) {
+  public Event(Organization organization, String eventTitle,String eventDateTime, String eventDescription, double eventLatitude, double eventLongitude, boolean foodAvaliable, boolean requiredFee) {
```
elefeint, line 44: Non-actionable musings follow.

At some point, might be that a no-argument constructor with setters for each field will be easier to manage than a constructor with lots of fields. Or... you could pass filters in a key-value map, and then set the individual fields in the constructor body.

But I am not sure how this will mesh with this being a data entity; needs to be tested and/or researched.

capstone/src/main/resources/static/event.js
```diff
@@ -1,9 +1,9 @@
 async function getEvents() {
   const response = await fetch('get-all-events');
   const jsonEvents = await response.json();
-  const eventList = document.getElementById('events-container');
+  //const eventList = document.getElementById('events-container');
```
elefeint, line 4: This page won't be showing events anymore? I guess it makes sense, since they are already searchable from the index page. Consider redirecting there from the event save then.

capstone/src/main/resources/static/event.js
elefeint: You might consider having a common library of useful utilities, shared across different pages.

capstone/src/main/java/com/step902020/capstone/Event.java
elefeint, line 44: I'd replace `boolean` with `Boolean`, because then old test data that did not have these fields can still be used.

But then you've got to check for `null` value and default to false instead; something like this:

```
this.foodAvaliable = foodAvaliable == null ? false : foodAvaliable;
this.requiredFee = requiredFee == null ? false : requiredFee;
```

oasemota99: I was actually beginning to look into this. I thought using setters would be better long term, but I'll dig a little deeper and see what I can find

capstone/src/main/resources/static/event.js
oasemota99: I will create an issue for it!

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -15,7 +15,6 @@
   @Id
   Long datastoreID;
 
-  @Reference
```
elefeint, line 18: Could you create a follow up issue to resolve the issue this was causing?

jsheng112: We came up with a solution to this problem. For the event side all the information that it needs from the organization is the organization id and the organization name. The plan is to directly store these as fields instead of as a reference and once we create the event element it will have a link to the public organization page with more info. A PR is coming soon to address this issue.

elefeint: Sounds good. It's a good example of the kind of architectural tradeoffs involved in building software!

PR #78, opened 2020-07-10
Format event listing

### LINKED ISSUE #38: Style event feed page

### LINKED ISSUE #64: Update event listing with more information

### LINKED ISSUE #71: Click on event/filter changes border color

### LINKED ISSUE #74: Organize & refactor CSS file

### LINKED ISSUE #75: 1st Iteration of standardizing design

lbourret:
-Create button to remove reviews (they're in temp location)
-Organize CSS file fixes #74 
-Standardize design elements  fixes #75
-Update event listing fixes #64 
-Style features fixes #71 fixes #38


capstone/src/main/resources/static/index.html
```diff
@@ -19,17 +20,13 @@
...
-      <label for="free">Free</label>
-      <input type="checkbox" id="free" name="free" value="free">
-      <label for="food">Food Provided</label>
-      <input type="checkbox" id="food" name="food" value="food">
-        
-        <label for="type">Event Type</label>
-        <select name="type" id="type">
-          <option disabled selected> -- select an option -- </option>
+      <button id="free" value="free" onclick="toggleBorderSelection('free')">FREE</button>
+      <button id="food" value="food" onclick="toggleBorderSelection('food')">FOOD</button>
```
jsheng112, line 26: These buttons look so good!

capstone/src/main/resources/static/profile.html
```diff
@@ -55,7 +56,7 @@
             <label>Last Name:</label>
             <input type="text" name="lastname" id="ind-lastname" placeholder="Your last name.."></input>
             <label>University:</label>
-            <p id="university-form-display"></p>
+            <p id="university-form-display" style="display:none"></p>
```
jsheng112, line 59: Correct me if I'm wrong, but I think this is the line you were talking about in my other PR. I think its fine if you hide it and then reveal it in the js function.

capstone/src/main/resources/static/script.js
```diff
@@ -250,6 +261,7 @@ function createIndividualProfile(data, fillForm) {
     // prefill form
     document.getElementById("ind-firstname").value = data.firstName;
     document.getElementById("ind-lastname").value = data.lastName;
+    document.getElementById("university-form-display").display = "block";
```
jsheng112, line 264: I see you revealed it in this function. No change on the functionality so we can keep it this way.

capstone/src/main/resources/static/script.js
```diff
@@ -425,6 +437,31 @@ function createCalendar() {
...
+}
 
+/*
+ * Reviews are temp on event listing so remove to see finalized event search page
+ */
+function removeReviews() {
+  const reviews = document.getElementsByClassName('reviews');
+  while(reviews.length > 0){
+    reviews[0].parentNode.removeChild(reviews[0]);
+  }
```
jsheng112, line 465: I like the layout of the events once the reviews are removed. Maybe this is a button we should add to temporarily hide reviews if the user does not care about them.

lbourret: Same! In the final project, I think the reviews will be attached to the event's detail page instead of the event search. Whenever those are created I'll just move the review element.

capstone/src/main/resources/static/script.js
```diff
@@ -425,6 +438,31 @@ function createCalendar() {
     dateDiv.appendChild(dateDisplay);
     calendar.append(dateDiv);
   }
+}
+/*
+ * If element is selected, border will change from white to green
+ * @param element id
+ */
+function toggleBorderSelection(elementId) {
```
elefeint: Optional: this could be simplified to *only* add/remove `selected` class. Then you don't need special treatment for button.

lbourret: Oh that's how I originally had it but I realized if a user selects another choice within select, then it will remove the class even though it was still "selected"

PR #80, opened 2020-07-10
Add identity backend

elefeint:
Changes in this PR:
* Authentication through IAP when app deployed.
* Authentication through Basic Authentication  in integration tests and when app run locally with `-P local`. Test account username is "student@uni.org"; test password is "test.
* Add REST GET endpoint `/identity` that returns the currently logged in user's e-mail
* Replace all X-identity headers with an injected component `CurrentUser`.
* Uncomment and fix tests
* For individual/organization saving, remove the "email" parameter in tests, since the latest behavior was to query the currently logged in user.

Unrelated changes:
* Added logback configuration file to help with troubleshooting.
* Replaced local `TestRestTemplate` with an injected instance that Spring Boot Test already configured with correct host/port.
  * As a result, removed concatenated `localhost + <RANDOM_PORT>` in tests.

Locally, you now have to run with a profile named "local": `mvn clean spring-boot:run -P local`.


capstone/pom.xml
```diff
@@ -34,9 +35,17 @@
     <dependency>
       <groupId>org.springframework.cloud</groupId>
       <artifactId>spring-cloud-gcp-starter-data-datastore</artifactId>
-      <version>1.2.3.RELEASE</version>
+      <version>${spring-cloud-gcp.version}</version>
```
meltsufin, line 38: We might as well bring in the BOM to avoid specifying versions for Spring Cloud GCP deps.

PR #94, opened 2020-07-14
Filtered search

lbourret:


capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -4,6 +4,8 @@
 import org.springframework.cloud.gcp.data.datastore.core.mapping.Field;
 import org.springframework.data.annotation.Reference;
 import org.springframework.data.annotation.Id;
+import org.springframework.lang.Nullable;
```
jsheng112, line 7: I maybe wrong but I don't think this was used anywhere. Is this the solution to the optional boolean issue?

capstone/src/test/java/com/step902020/capstone/UserDatastoreTest.java
```diff
@@ -31,7 +31,7 @@
     webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 public class UserDatastoreTest {
 
-  /* Properties for Test account from test/resources/application-local.properties */
+/*  *//* Properties for Test account from test/resources/application-local.properties *//*
```
meltsufin, line 34: Why did you comment this test out?

PR #95, opened 2020-07-14
replace session storage with server-side endpoint

jsheng112:
add a new endpoint to determine the user type.

### DIFF
--- a/capstone/src/main/java/com/step902020/capstone/EventController.java
+++ b/capstone/src/main/java/com/step902020/capstone/EventController.java
@@ -50,7 +50,7 @@ public class EventController {
      @RequestParam("requiredFee") Optional<Boolean> requiredFee,
      @RequestParam("event-id") String eventId
     ) throws IOException {
-      Organization organization = organizationRepository.findByEmail(email.substring(20)).get(0);
+      Organization organization = organizationRepository.findByEmail(email.substring(20)).orElse(null);
       Event event = eventId.length() <= 0? null : this.eventRepository.findById(Long.parseLong(eventId)).orElse(null);
       if (event != null) {
         event.setEventDateTime(eventDateTime);
--- a/capstone/src/main/java/com/step902020/capstone/EventController.java
+++ b/capstone/src/main/java/com/step902020/capstone/EventController.java
@@ -52,7 +52,7 @@ public class EventController {
      @RequestParam("requiredFee") Optional<Boolean> requiredFee,
      @RequestParam("event-id") String eventId
     ) throws IOException {
-      Organization organization = organizationRepository.findByEmail(user.getEmail()).get(0);
+      Organization organization = organizationRepository.findByEmail(user.getEmail()).orElse(null);
       Event event = eventId.length() <= 0? null : this.eventRepository.findById(Long.parseLong(eventId)).orElse(null);
       if (event != null) {
         event.setEventDateTime(eventDateTime);
--- a/capstone/src/main/java/com/step902020/capstone/security/IdentityController.java
+++ b/capstone/src/main/java/com/step902020/capstone/security/IdentityController.java
@@ -1,5 +1,10 @@
 package com.step902020.capstone.security;
 
+import com.step902020.capstone.Individual;
+import com.step902020.capstone.IndividualRepository;
+import com.step902020.capstone.Organization;
+import com.step902020.capstone.OrganizationRepository;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.core.context.SecurityContextHolder;
@@ -7,11 +12,35 @@ import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RestController;
 
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 @RestController
 public class IdentityController {
+
+  @Autowired
+  private IndividualRepository individualRepository;
+
+  @Autowired
+  private OrganizationRepository organizationRepository;
 
   @GetMapping("identity")
   public String getIdentity(CurrentUser user) {
     return user.getEmail();
+  }
+
+  @GetMapping("user-info")
+  public Object getUserType(CurrentUser user)  {
+    Organization organization = this.organizationRepository.findByEmail(user.getEmail()).orElse(null);
+    if (organization != null) {
+      return organization;
+    } else {
+      Individual individual = this.individualRepository.findByEmail(user.getEmail()).orElse(null);
+      if (individual != null) {
+        return individual;
+      }
+    }
+    return null;
   }
 }
--- a/capstone/src/main/java/com/step902020/capstone/security/IdentityController.java
+++ b/capstone/src/main/java/com/step902020/capstone/security/IdentityController.java
@@ -31,16 +31,23 @@ public class IdentityController {
   }
 
   @GetMapping("user-info")
-  public Object getUserType(CurrentUser user)  {
+  public Map<String, String> getUserType(CurrentUser user)  {
     Organization organization = this.organizationRepository.findByEmail(user.getEmail()).orElse(null);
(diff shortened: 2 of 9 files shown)


capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -52,18 +52,17 @@ public RedirectView saveEvent (
...
     ) throws IOException {
-      Organization organization = organizationRepository.findByEmail(user.getEmail()).get(0);
+      Organization organization = organizationRepository.findByEmail(user.getEmail()).orElse(null);
       Event event = eventId.length() <= 0? null : this.eventRepository.findById(Long.parseLong(eventId)).orElse(null);
       if (event != null) {
         event.setEventDateTime(eventDateTime);
         event.setEventDescription(eventDescription);
         event.setEventLatitude(Double.parseDouble(eventLatitude));
         event.setEventLongitude(Double.parseDouble(eventLongitude));
         event.setEventTitle(eventTitle);
```
oasemota99, line 62: (just a heads up) There could be a potential error here involving data retrieval and the hardcoded filters I added. I plan on looking into it but got lost with maps lol (there was something about the Optional<boolean> type and getting the filter that java didn't like). Likely something to look into sooner rather than later

oasemota99: If you want to see this locally, make setters for the hardcoded filters to see the compile error I got before

oasemota99: created issue #103; I'll fix it

jsheng112: I did not get any compile error with the setters, so there may not be an issue after all?

capstone/src/main/resources/static/user.js
```diff
@@ -172,38 +138,34 @@ function hideFields(selectField, universityField) {
...
-      data[0].savedEvents.forEach((event) => eventDiv.appendChild(createSavedEventElement(event, false, false)));
-      displayMain(true);
-    });
-  } else {
-    displayMain(false);
-  }
-
+  fetch('user-info').then(response => response.json()).then((data) => {
+    var userType = data.userType;
+    if(userType == "individual") {
```
meltsufin, line 143: It seems wasteful to do a call to `user-info` endpoint just to check that the current user is an individual. Since "individual" is in the name of the function, you can put the burden on the caller to make sure that they are only calling it for individuals.

capstone/src/main/resources/static/user.js
```diff
@@ -172,38 +138,34 @@ function hideFields(selectField, universityField) {
...
-      orgDiv.innerHTML='';
-      data[0].organizations.forEach((org) => orgDiv.appendChild(createSavedOrgElement(org, true, true)));
-      displayMain(true);
-    });
-  } else {
-    displayMain(false);
-  }
+  fetch('user-info').then(response => response.json()).then((data) => {
+    var userType = data.userType;
+    if (userType == "individual") {
```
meltsufin, line 159: Same comment as above.

capstone/src/main/resources/static/user.js
```diff
@@ -344,89 +305,87 @@ function closeForm() {
...
-      const eventDiv = document.getElementById("created-events");
-      data[0].events.forEach((event) => eventDiv.appendChild(createSavedEventElement(event, false, true)));
-      displayMain(true);
-    });
-  } else {
-    displayMain(false);
-  }
+  fetch('user-info').then(response => response.json()).then((data) => {
+    var userType = data.userType;
+    if (userType == "organization") {
```
meltsufin, line 310: Same as for "individual", I would avoid the roundtrip to the server just to ensure that the function is being invoked for an organization.

capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -82,12 +78,29 @@ public double getEventLongitude() {
...
+
+  public Boolean getFoodAvaliable() {
+    return foodAvaliable;
+  }
+
+  public Boolean getRequiredFee() {
+    return requiredFee;
+  }
+
+  public void setDatastoreID(Long datastoreID) {
```
elefeint, line 102: I'd suggest agreeing on a naming convention as a team about whether it's "Id" or "ID". Pure camel-case is more common in Java ("Id"), but it's more important to be consistent within a project.

capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -52,18 +52,17 @@ public RedirectView saveEvent (
      @RequestParam("requiredFee") Optional<Boolean> requiredFee,
      @RequestParam("event-id") String eventId
     ) throws IOException {
-      Organization organization = organizationRepository.findByEmail(user.getEmail()).get(0);
+      Organization organization = organizationRepository.findByEmail(user.getEmail()).orElse(null);
```
elefeint, line 55: See the previous PR for suggestion on simplifying this with a custom repository method name.

capstone/src/main/java/com/step902020/capstone/security/IdentityController.java
```diff
@@ -1,17 +1,49 @@
...
+    Organization organization = this.organizationRepository.findByEmail(user.getEmail()).orElse(null);
+    if (organization != null) {
+      return organization;
+    } else {
+      Individual individual = this.individualRepository.findByEmail(user.getEmail()).orElse(null);
+      if (individual != null) {
+        return individual;
+      } else {
+        Map<String, String> map = new HashMap<String, String>();
+        map.put("userType", "null");
```
elefeint, line 44: Because "null" has such a special meaning in Java and Javascript, I'd recommend using a custom type such as "unknown".

capstone/src/main/resources/static/script.js
```diff
@@ -17,25 +17,24 @@ var reviewsExist = false;
...
+        eventListElement.innerText = ""; // Clear elements in div
+
+        events.forEach((event) => {
+          eventListElement.appendChild(createEventElement(event, displaySaveButton));
+        })
+        // Format time to *** time ago
+        if (reviewsExist){
+          timeago.render(document.querySelectorAll('.timeago'));
+        }
+      });
```
elefeint, line 35: I'd move this block into a helper function to make the main (is user undefined? show main page. Is it an individual or an organization? fetch all events) logic easier to follow.

capstone/src/main/resources/static/user.js
```diff
@@ -1,26 +1,23 @@
...
+    } else {
+      if (data.userType == "organization") {
+        createProfile(data, fillForm, true);
         displayNavToggle("individual-nav", "org-nav");
-        sessionStorage.setItem("user-type", data[0].userType);
-      });
-    } else if (userType == "individual") {
-      fetch('get-individual').then(response => response.json()).then((data) => {
-        createProfile(data[0], fillForm, false);
+        sessionStorage.setItem("user-type", data.userType);
```
elefeint, line 12: These three lines are the same as in `else` clause; the only differences are the class tyle and true/false parameter in `createProfile`. So the whole if/else block could become one line invocation of a helper method. You don't even need to pass the `individual-nav`/`org-nav` class type -- the helper method can concatenate `userType` with `"-nav"`.

capstone/src/main/resources/static/user.js
```diff
@@ -172,38 +132,28 @@ function hideFields(selectField, universityField) {
...
+/* Function to get the saved organizations */
 function getIndividualOrganizations() {
-  var userType = sessionStorage.getItem("user-type");
-  if (userType == null) {
-    getUserType(true);
-  }
-  if (userType == "individual") {
-    fetch('get-' + userType).then(response => response.json()).then((data) => {
+  fetch('user-info').then(response => response.json()).then((data) => {
+    if(data.userType == "individual") {
```
elefeint, line 149: You know how we usually ask for more helper functions? In this case, I think the opposite could be helpful -- because `getIndividualEvents()` and `getIndividualOrganizations()` are really similar, perhaps they could merge to become a single helper method.

capstone/src/main/resources/static/script.js
jsheng112: After I merged with master this piece of code was already condensed (probably by Lian)

PR #102, opened 2020-07-16
Add hardcoded map markers

oasemota99:


capstone/src/main/resources/static/event.html
```diff
@@ -9,11 +9,16 @@
...
-        <a href="map.html"> Campus Map </a>
-      </nav>
-    </div>
+    <nav class="nav-bar">
+        <ul>
+            <li class="home"><a href="index.html">College Activity Finder</a></li>
+            <li class="active"><a href="index.html">Events</a></li>
+            <li><a href="organizationsearch.html">Organizations</a></li>
+            <li class=""><a href="profile.html">My Account</a></li>
+            <li class=""><a href="map.html">Event Map</a> </li>
```
jsheng112, line 18: Map looks great!

capstone/src/main/resources/static/index.html
```diff
@@ -17,6 +17,7 @@
         <li class="active"><a href="index.html">Events</a></li>
         <li><a href="organizationsearch.html">Organizations</a></li>
         <li class=""><a href="profile.html">My Account</a></li>
+        <li><a href="map.html">Event Map</a> </li>
```
jsheng112, line 20: It would truly be helpful to start using Thymeleaf. I am going to start working on it so that we don't have to manually change every navbar.

PR #104, opened 2020-07-16
Profile picture

jsheng112:
- implemented profile picture feature for user profile. If no profile picture was submitted a default image will be loaded.
- Cleaned up the objects created during tests.


capstone/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+# Uncomment this line and insert your own json key file for local testing
+spring.cloud.gcp.storage.credentials.location=file:/home/jennysheng/service-account.json
```
meltsufin, line 2: Wouldn't this configuration also cause the application to look for the file when deployed to App Engine?
And it won't find it.

capstone/pom.xml
```diff
@@ -81,6 +81,17 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-storage</artifactId>
+      <version>1.108.0</version>
+    </dependency>
```
elefeint, line 88: You should not need this dependency -- `spring-cloud-gcp-starter-storage` brings a compatible version transitively.

capstone/src/main/resources/application.properties
elefeint: +1; definitely comment it out when committing.

jsheng112:
On the topic of making it refresh, I have no idea why but the the form does not refresh like other forms do. I have to look into the issue more and will file an issue for it.

capstone/pom.xml
jsheng112: The policy is in this dependency so I still need to include it.

jsheng112:
fixes #107

capstone/pom.xml
meltsufin: @elefeint `PostPolicyV4` doesn't exist in the older version of `google-cloud-storage` that is pulled in transitively.

capstone/src/main/java/com/step902020/capstone/GcsStore.java
```diff
@@ -0,0 +1,74 @@
...
+  Storage storage;
+
+  /**
+   * generates a form for the user to submit profile picture
+   * @param projectId id of the current project
+   * @param bucketName bucket name preset to spring-bucket-jennysheng
+   * @param blobName name of the object matches user email
+   * @return html form in String
+   */
+  public String generateSignedPostPolicyV4(
```
meltsufin, line 24: I'm a bit confused by this approach of generating the whole HTML form and using this PolicyV4 stuff. Can you just use a signed URL to which you can `PUT` an image, like described [here](https://github.com/spring-cloud/spring-cloud-gcp/issues/411#issuecomment-364949224)?
You should also be able to just change the `action` attribute of the form in JS rather than returning the whole HTML snippet for the form from the server.

jsheng112: When I tried it without the PolicyV4 it gave an error saying that the identity of the user is anonymous. Also for the error that you got, can you tell me the size of the file that you tried to upload? I know this error exists and from what I can tell it seems like the error happens when the file is too small (smaller than 70kb, from what I can gather from testing various image sizes), but I could not figure out exactly what is causing the error. I will look into this more.

meltsufin: Never mind. You can't really do a `PUT` from a form, and it won't be a simple change.

capstone/src/main/java/com/step902020/capstone/GcsStore.java
```diff
@@ -0,0 +1,74 @@
...
+    for (Map.Entry<String, String> entry : policy.getFields().entrySet()) {
+      htmlForm.append(
+              "  <input name='"
+                      + entry.getKey()
+                      + "' value='"
+                      + entry.getValue()
+                      + "' type='hidden' />\n");
+    }
+    htmlForm.append("  <input type='file' name='file' required/><br />\n");
+    htmlForm.append("  <input type='submit' value='Upload File' name='submit' onsubmit='closeImageForm();'/><br />\n");
```
meltsufin, line 50: The error I had before was caused by the `submit` input field being processed on the backed. You just need to mark it to be ignored by prepending `x-ignore`. I figured this out by looking in the internal Google repository.
After this change, I was able to uploaded my image, which is only 6kb. Frankly, I'm not sure why images over 70Kb have been working for you.
```suggestion
    htmlForm.append("  <input type='submit' value='Upload File' name='x-ignore-submit' onsubmit='closeImageForm();'/><br />\n");
```

meltsufin:
Trying to fix the documentation at the source:
https://github.com/googleapis/google-cloud-java/pull/7292

capstone/src/main/java/com/step902020/capstone/GcsStore.java
jsheng112: Instead of committing this change I made the change in another commit.

jsheng112:
Yes the refresh issue is because there is no redirect back to the page so the page is not loaded again. I've been trying to add load functions to make the page load after submitting but nothing works. I created an issue for this.

jsheng112:
Refresh issue is fixed! The page will automatically refresh itself 10 seconds after submission.

capstone/src/main/java/com/step902020/capstone/OrganizationController.java
```diff
@@ -113,4 +113,15 @@ public RedirectView deleteOrganizationEvent(@RequestParam("event-id") String eve
     }
     return new RedirectView("manageevents.html", true);
   }
+
+  /**
+   * returns image with the same name as the user email from cloud storage
```
elefeint, line 118: That's excellent use of Javadoc -- to convey intent and let the user know what behavior to expect.

capstone/src/main/resources/static/user.js
```diff
@@ -39,6 +39,21 @@ function setUpAccountPage(isOrganization, fillForm, data, hide, display) {
...
+    const form = document.getElementById(formName);
+    form.innerHTML = data;
+  });
+
+}
+
+/* upon submission, hide the image form */
+function closeImageForm() {
+  document.getElementById("user-image-form").style.display = "none";
+  setTimeout(function () { window.location.reload(); }, 10000);
```
meltsufin, line 54: Why did you chose 10s? It seems pretty long.
Also, instead of refreshing the whole page, have you considered just forcing the image to refresh?
You can do it by setting `src` of the image.
Something like this:
```
document.getElementById("public-image-img").setAttribute("src", "get-public-image?email=" + data.email + new Date().getTime());
```

You can even do it multiple times for 10s or something like that.

capstone/src/main/resources/static/user.js
```diff
@@ -413,6 +432,8 @@ function getPublicProfile() {
    createProfile(data, false, true);
    const eventDiv = document.getElementById("hosted-events");
    data.events.forEach((event) => eventDiv.appendChild(createSavedEventElement(event, true, false)));
+   document.getElementById("public-image-a").setAttribute("href", "get-public-image?email=" + data.email);
```
meltsufin, line 435: So is the profile image only supported for organizations?
Or is the comment on this function incorrect?

jsheng112: Yes only public organizations will have a public profile for other users to view.

capstone/src/main/resources/static/user.js
jsheng112: The 10 seconds is arbitrary. On my computer it takes about 8 seconds so I just added a little more time than that to give some leeway.

PR #115, opened 2020-07-21
Event recommendation

### LINKED ISSUE #106: add tests to test delete event after PR #95 is merged

jsheng112:
- basic event recommendation system is set up. 
- There are only a few tests right now that have been added but I just want you guys to take a look at the algorithm so far before I continue to add more tests. 
- Feel free to test in the app as well through the new event recommendation tab.
- fixes #106


capstone/src/main/java/com/step902020/capstone/Event.java
```diff
@@ -134,4 +134,35 @@ public void addReview(Review review) {
...
+    return eventTitle + " " + datastoreId;
+  }
+
+  /**
+   * implement equality
+   * @param o object being compared to
+   * @return boolean
+   */
+  @Override
+  public boolean equals(Object o) {
```
lbourret, line 151: What is this used for?

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,68 @@
...
+      }
+    });
+
+    // put data from sorted list to hashmap
+    List<E> sorted = new ArrayList<>();
+    for (Map.Entry<U, Double> entry : tempList) {
+      List<E> currUserEvents = getItemList.apply(entry.getKey());
+      for (E e : currUserEvents) {
+        if (!targetUserEvents.contains(e) && !sorted.contains(e)) {
+          sorted.add(e);
```
lbourret, line 62: is there a limit to the list?

capstone/src/main/java/com/step902020/capstone/Event.java
jsheng112: To override the object equals. When the events are returned from the database they are not the same objects anymore but they should be considered "equal" if their ids are equal.

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,70 @@
...
+      }
+    });
+
+    // put data from sorted list to hashmap
+    List<E> sorted = new ArrayList<>();
+    int count = 0;
+    for (Map.Entry<U, Double> entry : tempList) {
+      List<E> currUserEvents = getItemList.apply(entry.getKey());
+      for (E e : currUserEvents) {
+        if (!targetUserEvents.contains(e) && !sorted.contains(e) && count < 10) {
```
meltsufin, line 62: Another instance of the magic number `10` that needs to be replace with a variable.

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,70 @@
...
+        if (!targetUserEvents.contains(e)) {
+          dist++;
+        }
+      }
+      double scaledDist = 1/(1+Math.sqrt(dist));
+      userToScore.put(user, scaledDist);
+    }
+    // sort the resulting list of individuals by increasing distance
+    // Create a list from elements of HashMap
+    List<Map.Entry<U, Double>> tempList =
```
meltsufin, line 43: Maybe rename `tempList` to something more descriptive, like `sortedUsers`.

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,70 @@
...
+    Collections.sort(tempList, new Comparator<Map.Entry<U, Double> >() {
+      @Override
+      public int compare(Map.Entry<U, Double> o1,
+                         Map.Entry<U, Double> o2)
+      {
+        return (o1.getValue()).compareTo(o2.getValue());
+      }
+    });
+
+    // put data from sorted list to hashmap
```
meltsufin, line 56: Did you mean to say "....to list"?

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,70 @@
...
+        if (!userEvents.contains(e)) {
+          dist++;
+        }
+      }
+      for (E e : userEvents) {
+        if (!targetUserEvents.contains(e)) {
+          dist++;
+        }
+      }
+      double scaledDist = 1/(1+Math.sqrt(dist));
```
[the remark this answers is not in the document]
meltsufin: Giving it a bit more thought, I think your distance calculation is actually correct and equivalent to the euclidian distance. However, I'm still not sure about the whole scaling part.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -170,4 +170,30 @@ public RedirectView deleteSavedOrganization(
...
+   * @param currentUser user that is currently logged in
+   * @return list of Events
+   */
+  @GetMapping("get-recommended-events-individual")
+  public List<Event> recommendEvents(CurrentUser currentUser) {
+    Individual targetUser = this.individualRepository.findFirstByEmail(currentUser.getEmail());
+    List<Individual> users = this.individualRepository.findByUniversity(targetUser.getUniversity());
+    List<Event> recommended = Recommender.recommend(targetUser, users, u -> u.getSavedEvents());
+    if (recommended.size() < 10) {
+      List<Event> allEvents = this.eventRepository.findAllByUniversity(targetUser.getUniversity());
```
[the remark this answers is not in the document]
jsheng112: I did try to do it like that, but the issue is I would potentially have duplicate events in the list. So sometimes I need to consider more than 10 events in order to fill in the gap.

capstone/src/main/java/com/step902020/capstone/Recommender.java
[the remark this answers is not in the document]
jsheng112: I agree the scaling part would not actually make a difference. However, I do have some concerns about the euclidean distance. Currently what I am doing is considering only events that either one or the other user have added. This means that for each user I do not consider their distance with respect to the entire event space. 
Example: 
- the list of all events: A, B, C, D
- user 1 saves A, B
- user 2 saves A, C
in this case the dimension would be 3 instead of 4 because in my distance calculation I only consider A, B, C
Another example:
- user 1 saves A,
- user 2 saves A, B
in this case the dimension I'm considering would be 2 because I only consider distance WRT to A, B

Is this a valid way of doing it? The differing dimensions are concerning me

capstone/src/main/java/com/step902020/capstone/EventRepository.java
```diff
@@ -1,9 +1,13 @@
...
+
+import java.util.List;
+
 /**
  * Spring Data Repository for Event Entities
  */
 public interface EventRepository extends DatastoreRepository<Event, Long> {
-
+  public long deleteAllByEventDateTime(String eventDateTime);
+  public List<Event> findAllByUniversity(String university);
```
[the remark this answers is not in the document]
jsheng112: I think they do the same thing, but for consistency its better to change it to findByUniversity.

capstone/src/main/java/com/step902020/capstone/Recommender.java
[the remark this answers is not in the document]
meltsufin: Well I think the scaling actually does make a difference because it inverts the values. If you start out with a distance between u1 & u2 as 100 and u1 & u3 as 1000, the scaling makes the distance between u1 & u2 as 1/sqrt(100)=1/10=**.1** and between u1 & u3 as 1/sqrt(1000)=1/100=**.01**. So, the smaller distance became the bigger distance.

In your first example:
u1 = [1, 1, 0, 0]
u2 = [1, 0, 1, 0]
d(u1, u2) = sqrt((1-1)^2 + (1-0)^2 + (0-1)^2 + (0-0)^2) = sqrt(0 + 1 + 1 + 0) = sqrt(2)

Second example:
u1 = [1, 1, 0, 0]
u2 = [1, 0, 0, 0]
d(u1, u2) = sqrt((1-1)^2 + (1-0)^2 + (0-0)^2 + (0-0)^2) = sqrt(0 + 1 + 0 + 0) = sqrt(1)

As you can, see you don't need to worry about events that exist but neither u1 or u2 save because they become 0s in the formula. The only events you need to add up are the ones where u1 saved and u2 didn't and vice versa, which is exactly what your algorithm is doing as far as I understand. BTW, the final sqrt doesn't really matter because it doesn't change the relative ranking.

capstone/src/main/java/com/step902020/capstone/Recommender.java
jsheng112: Not at the moment, but you are right adding a limit would be nice

capstone/src/main/java/com/step902020/capstone/Recommender.java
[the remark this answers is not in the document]
jsheng112: Ohh yes I see. In this case I won't use the scaling and will just sort in ascending order.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -179,6 +179,71 @@ public RedirectView deleteSavedOrganization(
...
+    int num = (count.equals("All")) ? Integer.MAX_VALUE : Integer.parseInt(count);
+    // find recommended events from other users
+    List<Event> recommended = Recommender.recommend(targetUser, users, u -> u.getSavedEvents(), num);
+    // if the list of recommended events is shorter than the list we want, add in more events from general event pool
+    if (recommended.size() < num) {
+      List<Event> allEvents = this.eventRepository.findByUniversity(targetUser.getUniversity());
+      int i = 0;
+      int length = 0;
+      int targetSize = recommended.size();
+      while (i < allEvents.size() && length < num - targetSize) {
```
elefeint, line 203: I'd make a temporary variable `int numExtraEvents =  num - targetSize` for ease of reading. The compiler will inline the definition anyway.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -179,6 +179,71 @@ public RedirectView deleteSavedOrganization(
...
+    List<Individual> users = this.individualRepository.findByUniversity(targetUser.getUniversity());
+
+    int num = (count.equals("All")) ? Integer.MAX_VALUE : Integer.parseInt(count);
+    // find recommended events from other users
+    List<Event> recommended = Recommender.recommend(targetUser, users, u -> u.getSavedEvents(), num);
+    // if the list of recommended events is shorter than the list we want, add in more events from general event pool
+    if (recommended.size() < num) {
+      List<Event> allEvents = this.eventRepository.findByUniversity(targetUser.getUniversity());
+      int i = 0;
+      int length = 0;
```
elefeint, line 201: I'd rename `length` to something more meaningful, like `numExtraEventsFound`.

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.servlet.view.RedirectView;
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.function.Function;
+
+public class Recommender {
+
+  public static <E, U> List<E> recommend (U targetUser, List<U> users, Function<U, List<E>> getItemList, int numNeeded) {
```
elefeint, line 21: Definitely add Javadoc documenting what the generic types `E` and `U` are expected to be.

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,73 @@
...
+      for (E e : targetUserEvents) {
+        if (!userEvents.contains(e)) {
+          dist++;
+        }
+      }
+      for (E e : userEvents) {
+        if (!targetUserEvents.contains(e)) {
+          dist++;
+        }
+      }
```
elefeint, line 41: How about a `calculateDistance()` helper method?

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,73 @@
...
+      userToScore.put(user, dist);
+    }
+    // sort the resulting list of individuals by increasing distance
+    // Create a list from elements of HashMap
+    List<Map.Entry<U, Integer>> sortedUsers =
+            new LinkedList<Map.Entry<U, Integer> >(userToScore.entrySet());
+
+    // Sort the list
+    Collections.sort(sortedUsers, new Comparator<Map.Entry<U, Integer> >() {
```
elefeint, line 50: Try lambdas here! This whole comparator can be expressed as `(o1, o2) -> (o1.getValue()).compareTo(o2.getValue())`.

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -0,0 +1,73 @@
...
+      for (E e : userEvents) {
+        if (!targetUserEvents.contains(e)) {
+          dist++;
+        }
+      }
+      userToScore.put(user, dist);
+    }
+    // sort the resulting list of individuals by increasing distance
+    // Create a list from elements of HashMap
+    List<Map.Entry<U, Integer>> sortedUsers =
```
elefeint, line 46: You might be interested in a `TreeMap` with a custom comparator. It's a sorted map; normally sorted by key but can also take a custom comparator, such as the one below.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -179,6 +179,71 @@ public RedirectView deleteSavedOrganization(
...
+                                @RequestParam("count") String count) {
+    Individual targetUser = this.individualRepository.findFirstByEmail(currentUser.getEmail());
+    List<Individual> users = this.individualRepository.findByUniversity(targetUser.getUniversity());
+
+    int num = (count.equals("All")) ? Integer.MAX_VALUE : Integer.parseInt(count);
+    // find recommended events from other users
+    List<Event> recommended = Recommender.recommend(targetUser, users, u -> u.getSavedEvents(), num);
+    // if the list of recommended events is shorter than the list we want, add in more events from general event pool
+    if (recommended.size() < num) {
+      List<Event> allEvents = this.eventRepository.findByUniversity(targetUser.getUniversity());
```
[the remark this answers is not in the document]
jsheng112: This is a very good point. I totally missed the issue with past events.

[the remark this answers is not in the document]
jsheng112: It turns out that we did not address past events in several other cases as well. I will address all of them in a separate PR.

capstone/src/main/java/com/step902020/capstone/Recommender.java
jsheng112: Correct me if I'm wrong, but I think TreeMap only keeps the keys in a sorted order, not values.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -179,6 +179,71 @@ public RedirectView deleteSavedOrganization(
...
+   */
+  @GetMapping("get-recommended-events-individual")
+  public List<Event> recommendEvents(CurrentUser currentUser,
+                                @RequestParam("count") String count) {
+    Individual targetUser = this.individualRepository.findFirstByEmail(currentUser.getEmail());
+    List<Individual> users = this.individualRepository.findByUniversity(targetUser.getUniversity());
+
+    int num = (count.equals("All")) ? Integer.MAX_VALUE : Integer.parseInt(count);
+    // find recommended events from other users
+    List<Event> recommended = Recommender.recommend(targetUser, users, u -> u.getSavedEvents(), num);
```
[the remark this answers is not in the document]
jsheng112: I am kind of confused by "creating a Recommender object at the top level, and allowing to override it with a setter." What does overriding it with a setter mean?

capstone/src/main/java/com/step902020/capstone/Recommender.java
elefeint: By default, yes. But you can convince it otherwise with the same comparator you had before.

elefeint: Here is an example:
```
  @Test
  public void treeMapSortedByValue() {
    Map<String, Integer> unsortedMap = new HashMap<>();
    unsortedMap.put("z", 6);
    unsortedMap.put("b", 4);
    unsortedMap.put("t", 7);
    unsortedMap.put("e", 2);
    unsortedMap.put("a", 9);
    unsortedMap.put("g", 1);
    unsortedMap.put("w", 5);


    Map<String, Integer> treeMapDefaultSort = new TreeMap<>(unsortedMap);
    Comparator<String> valueComparator = (s1, s2) -> unsortedMap.get(s1) - unsortedMap.get(s2);
    Map<String, Integer> treeMapValueSort = new TreeMap<String, Integer>(valueComparator);
    treeMapValueSort.putAll(unsortedMap);

    // "a", "b", "e", "g", "t", "w", "z"
    assertArrayEquals(new Integer[] {9, 4, 2, 1, 7, 5, 6},
      treeMapDefaultSort.values().toArray(new Integer[0]));

    assertArrayEquals(new Integer[] {1, 2, 4, 5, 6, 7, 9},
        treeMapValueSort.values().toArray(new Integer[0]));

  }
```

capstone/src/main/java/com/step902020/capstone/IndividualController.java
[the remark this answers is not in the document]
elefeint: you can unit test `IndividualController` by injecting a different version of recommender, like this:
```
Recommender mockRecommender = Mockito.mock(Recommender.class);
when(mockRecommender.recommend(...)).thenReturn(/*hardcoded recommendations*/)
IndividualController unitUnderTest = new IndividualController();
unitUnderTest.setRecommender(mockRecommender);
```

PR #116, opened 2020-07-21
Update review functionality

### LINKED ISSUE #59: Add ability to "like" reviews

### LINKED ISSUE #112: User can edit & delete their review

### LINKED ISSUE #120: Standardize button hover behavior

### LINKED ISSUE #128: Onload events in index.html

lbourret:
-Users can like review once fixes #59
-Only individuals can submit a review
-Authors of reviews can edit and delete their reviews fixes #112 
-UX changes fixes #128 fixes #120

### DIFF
--- a/capstone/src/main/java/com/step902020/capstone/EventController.java
+++ b/capstone/src/main/java/com/step902020/capstone/EventController.java
@@ -32,6 +32,9 @@ public class EventController {
   private EventRepository eventRepository;
 
   @Autowired
+  private IndividualRepository individualRepository;
+
+  @Autowired
   private OrganizationRepository organizationRepository;
 
   @Autowired
@@ -102,14 +105,24 @@ public class EventController {
       return new RedirectView("manageevents.html", true);
   }
 
+  /**
+   * Add new review to event
+   * @param user current user
+   * @param eventId Event's datastore id
+   * @param text Review's text
+   * @return Updated review list
+   */
   @PostMapping("/new-review")
   public List<Review> addReview(
+          CurrentUser user,
           @RequestParam("text") String text,
-          @RequestParam("eventId") Long eventId,
-          @RequestParam("name") String name) throws IOException {
+          @RequestParam("eventId") Long eventId) throws IOException {
 
     Event event = this.eventRepository.findById(eventId).get();
-    Review review = new Review(text, name);
+    Individual individual = this.individualRepository.findFirstByEmail(user.getEmail());
+    String individualName = individual.firstName + " " + individual.lastName;
+    String individualEmail = individual.email;
+    Review review = new Review(individualName, individualEmail, text);
     event.addReview(review);
     this.eventRepository.save(event);
     return event.reviews;
--- a/capstone/src/main/java/com/step902020/capstone/Review.java
+++ b/capstone/src/main/java/com/step902020/capstone/Review.java
@@ -1,11 +1,18 @@
 package com.step902020.capstone;
 
+import com.fasterxml.jackson.annotation.JsonBackReference;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.Entity;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.Field;
 import org.springframework.data.annotation.Id;
+import org.springframework.data.annotation.Reference;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
- * 
+ * Represents a Review. A Review is attached to en Event
+ * and contains the user who created  it, the text content, and the amt of likes it has received
+ * @author lbourret
  */
 @Entity(name = "review")
 public class Review {
@@ -14,32 +21,99 @@ public class Review {
 
   long timestamp;
 
-  @Field(name="name")
-  String name;
+  int likes;
+
+  String individualName;
+
+  String individualEmail;
 
   @Field(name="text")
   String text;
 
+  List<String> reviewLikers;
+
   /**
-   * Create a new Review
+   * Creates a new Review
+   * @param individualName Name of individual who authored review
    * @param text Text content of review
-   * @param name Review's author
    */
-  public Review(String text, String name) {
+  public Review(String individualName, String individualEmail, String text) {
     this.timestamp = System.currentTimeMillis();
+    this.likes = 0;
+    this.individualName = individualName;
+    this.individualEmail = individualEmail;
     this.text = text;
-    this.name = name;
+    this.reviewLikers = new ArrayList();
   }
 
-  public String getName() {
-    return name;
+  public Long getDatastoreId() {
+    return datastoreId;
   }
 
   public long getTimestamp() {
     return timestamp;
   }
 
+  public int getLikes() {
+    return likes;
+  }
+
+  public String getIndividualName() {
+    return individualName;
+  }
+
+  public String getIndividualEmail() {
+    return individualEmail;
+  }
+
   public String getText() {
     return text;
+  }
+
+  public List<String> getReviewLikers() {
+    return reviewLikers;
+  }
+
+  public void setText(String text) {
+    this.text = text;
+  }
+
+  /**
+   * Increase review's like count by one
+   */
+  public void addLike() {
+    likes++;
+  }
+
+  /**
+   * Decrease review's like count by one
+   */
+  public void removeLike() {
+    likes--;
+  }
+
+  /**
+   * Remove individual who added their like
+   * @param email Email of individual who liked review
+   */
+  public void addReviewLiker(String email) {
+    reviewLikers.add(email);
+  }
+
+  /**
+   * Remove individual who removed their like
+   * @param email Email of individual who unlike review
+   */
+  public void removeReviewLiker(String email) {
+    reviewLikers.remove(email);
+  }
+
+  /**
+   * Individual has liked this review
+   * @param email Email of Individual to check if they've liked review
+   * @return true if individual has liked review
+   */
+  public boolean isReviewLiker(String email) {
+    return reviewLikers.contains(email);
   }
 }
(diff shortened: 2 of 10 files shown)


lbourret:
For some of my features, I update the DOM without waiting for the backend to complete so there's a discrepancy if you like/delete/edit a review and immediately click out and back in. I wasn't sure if this was worth going back to fix since I'd imagine most users wouldn't immediately click out and back, but it's that's a problem I can file an issue.

capstone/src/main/resources/static/event.js
```diff
@@ -28,108 +29,188 @@ function getAllEventsForSearch(data) {
...
   })
 }
 
+/**
+ * Add delete and edit functionality for review's author
+ * @param reviewContainer Review's container
+ * @param reviewTextElement Container for review's text
+ * @param reviewId Review's datastore id
+ */
+function reviewAuthorFeatures(reviewContainer, reviewTextElement, reviewId) {
```
jsheng112: great features to have!

capstone/src/main/resources/static/event.js
```diff
@@ -28,108 +29,188 @@ function getAllEventsForSearch(data) {
...
   getEvents();
   return reviews;
+}
 
+/**
+ * Remove a review from event's list
+ * @param reviewId Review to remove's datastoreId
+ */
+function deleteReview(reviewId) {
+  const params = new URLSearchParams();
```
jsheng112, line 281: I like how you appended new params from the search params.

capstone/src/main/resources/static/event.js
```diff
@@ -28,108 +29,188 @@ function getAllEventsForSearch(data) {
...
+ * Add to review's like count
+ * @param reviewId Review's datastoreId
+ * @return Updated like count
+ */
+async function likeReview(reviewId) {
+  const params = new URLSearchParams();
+  params.append('reviewId', reviewId);
+  const response = await fetch('review-likes', {method:'POST', body: params});
+  const value = response.json();
+  getEvents();
```
jsheng112, line 297: Instead of calling getEvents for all these functions, I think you can move this function to be called when the body is onload. This way the user would see a list of events when they enter the index.html and you also don't need to call it for every function.

lbourret: oh yeah that's a good idea! #128

capstone/src/main/java/com/step902020/capstone/Review.java
```diff
@@ -14,32 +21,99 @@
 
   long timestamp;
 
-  @Field(name="name")
-  String name;
+  int likes;
+
+  String individualName;
+
+  String individualEmail;
```
elefeint, line 28: Should those be fields?

capstone/src/main/java/com/step902020/capstone/ReviewController.java
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+      review.addLike();
+      review.addReviewLiker(email);
+    }
+    this.reviewRepository.save(review);
+    return review.likes;
+  }
+
+  /**
+   * Delete review (Only review's author can delete review)
```
elefeint, line 52: So this says! But there is no backend check for current user, and frontend checks are merely a "do not tresspass" sign, so... *puts hackerz hat on*

capstone/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+# Uncomment this line and insert your own json key file for local testing
+//spring.cloud.gcp.storage.credentials.location=file:/home/lbourret/service_account.json
```
elefeint, line 2: Does `//` actually work to comment out properties files' lines?

capstone/src/main/resources/static/event.js
```diff
@@ -1,25 +1,72 @@
+/* Function to prefill event information if editing event */
+function loadEventInfo() {
+  const event = window.location.hash.substring(1);
```
elefeint, line 3: I left the same comment on Jenny's PR -- it is unusual to use the anchor tag (`#section`) to convey variables' values. It's more typical to use the query string `?key=value&key2=value2` for this.
When using the query string, you can use URLSearchParams to extract individual values.

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

capstone/src/main/java/com/step902020/capstone/Review.java
```diff
@@ -14,32 +21,99 @@
...
    */
-  public Review(String text, String name) {
+  public Review(String individualName, String individualEmail, String text) {
     this.timestamp = System.currentTimeMillis();
+    this.likes = 0;
+    this.individualName = individualName;
+    this.individualEmail = individualEmail;
     this.text = text;
-    this.name = name;
+    this.reviewLikers = new ArrayList();
```
oasemota99, line 46: are you planning on displaying the people who like a review or is it just to keep track of whether or not a user has liked a review?

lbourret: Right now, it's to keep track so a user can't like more than once

capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -66,22 +69,34 @@ public RedirectView saveEvent (
...
       }
       return new RedirectView("manageevents.html", true);
   }
 
+  /**
+   * Add new review to event
+   * @param user current user
+   * @param eventId Event's datastore id
+   * @param text Review's text
+   * @return Updated review list
```
elefeint, line 87: Extreme nit: usually field description starts with lowercase.

capstone/src/main/java/com/step902020/capstone/ReviewController.java
```diff
@@ -0,0 +1,83 @@
...
+  private IndividualRepository individualRepository;
+
+  /**
+   * Toggle user's like. Add like if they have not already liked it. Remove like if they have already liked it
+   * @param user current user
+   * @param reviewId review's datastore id
+   * @return amount of review's likes
+   */
+  @PostMapping("review-likes")
+  public int reviewLikes(
```
elefeint, line 34: Consider naming the method with an action, such as `updateReviewLikes()`.
The URL could go either way (stay as is or change to action). The pure REST convention is to have a noun as URL, and use the method (such as POST) to convey action. But you see plenty of action verbs in URLs, too, because they are easy for people to immediately understand.

capstone/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+# Uncomment this line and insert your own json key file for local testing
+spring.cloud.gcp.storage.credentials.location=file:/home/lbourret/service_account.json
```
elefeint, line 2: Remember to comment this out when committing.

capstone/src/main/resources/static/event.js
```diff
@@ -1,25 +1,72 @@
...
+ * @param campusMap - Google Map object
+ */
+function createMarker(event,campusMap) {
+  var eventPosition = {lat: event.eventLatitude, lng: event.eventLongitude};
+  const newMarker = new google.maps.Marker({
+    map: campusMap,
+    title: event.eventTitle,
+    position: eventPosition
+  })
+}
```
elefeint, line 45: This might have dropped in from another PR.

capstone/src/main/resources/static/event.js
```diff
@@ -28,167 +75,240 @@ function getAllEventsForSearch(data) {
...
+
+    const reviewLikeElement = createElement(reviewContainer, 'button',  review.likes + ' Likes');
+    reviewLikeElement.addEventListener('click', () => {
+      likeReview(review.datastoreId).then((reviewLikes) => {
+        reviewLikeElement.innerText = reviewLikes + ' Likes';
+      });
+    });
+
+    if (review.individualEmail == userEmail) {
+      reviewDeleteButton(reviewContainer, review.datastoreId);
```
elefeint, line 220: Consider naming function as an action - `showReviewDeleteButton()`.

capstone/src/main/resources/static/event.js
```diff
@@ -28,167 +75,240 @@ function getAllEventsForSearch(data) {
...
+ * @param reviewContainer Review's container
+ * @param reviewTextElement Container for review's text
+ * @param reviewId Review's datastore id
+ */
+function reviewEditButton(reviewContainer, reviewTextElement, reviewId) {
+  const editButton = createElement(reviewContainer, 'button', 'Edit');
+  editButton.addEventListener('click', () => {
+    if (editButton.innerText == 'Edit') {
+      reviewTextElement.contentEditable = true;
+      reviewTextElement.focus();
```
elefeint, line 250: Nice!

capstone/src/main/java/com/step902020/capstone/Review.java
meltsufin: You only need to mark something as `@Field` when you want to change the default `name`. BTW, `text` field below doesn't need `@Field`.

capstone/src/main/java/com/step902020/capstone/Review.java
```diff
@@ -14,32 +21,99 @@
 
   long timestamp;
 
-  @Field(name="name")
-  String name;
+  int likes;
+
+  String individualName;
```
meltsufin, line 26: Have you considered just using a reference to `Individual`?

lbourret: Yes that's what I had originally, but if you leave a review on an event that you've also saved, you get an infinite recursion error

capstone/src/main/resources/static/event.js
lbourret: Oh I just moved some of the functions around so they'd be grouped by functionality!

capstone/src/main/java/com/step902020/capstone/Review.java
meltsufin: I see. This must be happening with JSON serialization. I think we need to come up with a general strategy for dealing with cycles.
Have you looked at `@JsonIdentityInfo`. That might solve the problem with cycles.

PR #129, opened 2020-07-23
add marker to mini map in event.html

### LINKED ISSUE #54: Add mini map or geolocation
allows users to place a marker on a map without using coordinates

oasemota99:
allows user to place marker on mini map in event.html using a single click (be sure NOT to double click)
fixes #54 
*** NOTE ***
This PR simply adds the functionality of creating a marker on a mini map. I don't recommend extensive testing of it. It certainly does not fully replace the current maps feature using coordinates.


capstone/src/main/resources/static/event.js
```diff
@@ -56,7 +56,7 @@ function createEventElement(eventListElement, event, displaySaveButton) {
 
   // Displays only for individual users
   if (displaySaveButton) {
-    eventElement.appendChild(createSaveEventButton(event));
```
elefeint, line 59: Does this change belong in this PR? Looks like something that might have dropped in from another one.

capstone/src/main/resources/static/map.js
```diff
@@ -0,0 +1,56 @@
...
+     google.maps.event.addListener(campusMap, 'click', function(event) {
+      placeMarkerAndPan(event.latLng, campusMap);
+     });
+  }
+
+/* Create a new marker for each event
+ * @param event - event object
+ * @param campusMap - Google Map object
+ */
+function createMarker(event,campusMap) {
```
elefeint, line 34: Really great naming and helper functions. Makes it easy to read the main logic.

capstone/src/main/resources/static/map.js
```diff
@@ -0,0 +1,56 @@
+/* Function to create Google Map in map.html page */
+async function createMap() {
+  var princetonLatLng = {lat: 40.3428452, lng: -74.6568153};
+  const campusMap = new google.maps.Map(
+    document.getElementById('map'),
+    {center: princetonLatLng, zoom: 16});
+
+   google.maps.event.addListener(campusMap, 'click', function(event) {
```
meltsufin, line 8: You probably don't need this functionality on the main events map.

capstone/src/main/resources/static/map.js
```diff
@@ -0,0 +1,56 @@
...
+
+/* Function to create Mini Map in event.html page */
+  function createEventPlacementMap() {
+    var princetonLatLng = {lat: 40.3428452, lng: -74.6568153};
+    const campusMap = new google.maps.Map(
+      document.getElementById('eventMap'),
+      {center: princetonLatLng, zoom: 16});
+
+     google.maps.event.addListener(campusMap, 'click', function(event) {
+      placeMarkerAndPan(event.latLng, campusMap);
```
meltsufin, line 26: I guess you're still working on populating the lat and long in the event form based on the click, which would go right after this line.

capstone/src/main/resources/static/event.js
oasemota99: I have no idea where this came from. I'm reverting the change

capstone/src/main/resources/static/map.js
oasemota99: resolved!

PR #133, opened 2020-07-24
Loading wheel

### LINKED ISSUE #91: screen refresh spinner during loading

jsheng112:
- add loading wheel to every page except map
fixes #91


capstone/src/main/resources/static/user.js
```diff
@@ -430,13 +440,27 @@ function createCalendarEventElement(event, eventTime) {
...
    data.events.forEach((event) => eventDiv.appendChild(createSavedEventElement(event, true, false)));
    document.getElementById("public-image-a").setAttribute("href", "get-public-image?email=" + data.email);
    document.getElementById("public-image-img").setAttribute("src", "get-public-image?email=" + data.email);
+  hideSpinner();
  });
 }
 
+/* displays spinner on the page and hide main content */
+function showSpinner() {
+  document.getElementById('contents').style.display="none";
```
meltsufin, line 457: Instead of hiding all of the main page content while the loading is in progress, can you just make the spinner float on top of the existing content using something like CSS `position: absolute`?
I think this would reduce the feeling of "flicker" when the page is blank while loading is happening.

capstone/src/main/resources/static/event.js
```diff
@@ -2,12 +2,14 @@
...
 function getEvents() {
+  showSpinner();
   fetch('user-info').then(response => response.json()).then((data) => {
     if (data.userType != "unknown") {
       getAllEventsForSearch(data);
     } else {
       displayMain(false);
     }
+    hideSpinner();
   });
```
elefeint, line 13: Also remove the spinner on error:
```
.catch((error) => {
  // log error
  hideSpinner();
});
```

Example use here: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Uploading_JSON_data

PR #136, opened 2020-07-28
add info-windows to maps

oasemota99:
Adds small info window to map markers on main map page. Click on the marker to see the name of the event
fixes basic functionality of #101


capstone/src/main/resources/static/map.js
```diff
@@ -5,7 +5,7 @@ async function createMap() {
     document.getElementById('map'),
     {center: princetonLatLng, zoom: 16});
 
-   const response = await fetch('get-all-events');
+   const response = await fetch('get-map-events');
```
lbourret, line 8: Oh in one of my prs I created a new mapping function for filtered events instead and left all-events as normal bc of the error it was causing on maps. So I think whatever pr gets pushed in first, we can just adjust to those mapping functions?

capstone/src/main/resources/static/map.js
```diff
@@ -40,8 +39,28 @@ function createMarker(event,campusMap) {
   const newMarker = new google.maps.Marker({
     map: campusMap,
     title: event.eventTitle,
-    position: eventPosition
+    position: eventPosition,
   });
+  var eventInfoWindow = createInfoWindow(event, campusMap);
+  var eventContent = '<p id=mapContent>'+ event.eventTitle + '</p>';
```
meltsufin, line 45: How can the user save the event from this view? Would it be possible to provide a link to the event details page? I'm not sure we have a separate event details page though, just a popup.

capstone/src/main/resources/static/map.js
```diff
@@ -40,8 +39,28 @@ function createMarker(event,campusMap) {
...
+  newMarker.addListener('click', function() {
+    eventInfoWindow.setContent(eventContent);
```
meltsufin, line 47: Would it work to just set the content in the `createInfoWindow()`, and only do the `open()` on the click event?

capstone/src/main/resources/static/map.js
jsheng112: So none of you guys use 'get-all-events' anymore? I have a PR that filters out past events for that end point but if you guys are not using it I'll just wait for these functions to be merged.

capstone/src/main/resources/static/map.js
oasemota99: There isn't a dedicated page, it's just a popup. I'm not sure if I could make a link to it or not. Also the events on the map are the saved events from the user

capstone/src/main/resources/static/map.js
oasemota99: I think filtering out the past events is really useful, so I'm fine with using that instead

PR #145, opened 2020-07-29
Add Rank to event and organizations

### LINKED ISSUE #127: Fix modal cutoff

### LINKED ISSUE #132: fix reference issue and circular serialization

jsheng112:
interestedUserCount is determined by the number of users who save/unsave the event or organization.
Change organization search to display in descending rank order. Event search untouched because Lian is working alot on that feature
adjust modal display to display rank
fixes #132 and fixes #127


capstone/src/main/resources/static/map.js
```diff
@@ -11,6 +11,15 @@ async function createMap() {
    jsonEvents.forEach(event => createMarker(event, campusMap));
   }
 
+function createMapForASingleEvent(event) {
+  var princetonLatLng = {lat: 40.3428452, lng: -74.6568153};
```
oasemota99, line 15: you can use the code I pushed to show the user's selected university instead

capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -74,12 +73,16 @@
         noPastEvents.add(e);
       }
     }
+
+    // order by rank
+    Collections.sort(noPastEvents, (a, b) -> Integer.compare(b.getRank(), a.getRank()));
```
meltsufin, line 78: Is it possible to rely on Datastore to do the sort when you query for events above?

capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -74,12 +73,16 @@
...
     return noPastEvents;
```
lbourret, line 79: It seems like the events don't print in rank order on the index page. I tried refreshing and the events would display in different orders each time and close, but not exact rank order

capstone/src/main/java/com/step902020/capstone/Organization.java
```diff
@@ -84,6 +90,20 @@ public void addEvent(Event event) {
     events.add(event);
   }
 
+  /**
+   * increase rank by 1
+   */
+  public void addRank() {
```
meltsufin, line 96: `incrementRank()` or `increaseRank(int)` maybe?

capstone/src/main/java/com/step902020/capstone/Organization.java
```diff
@@ -84,6 +90,20 @@ public void addEvent(Event event) {
...
+    rank++;
+  }
+
+  /**
+   * decrease rank by 1
+   */
+  public void minusRank() {
```
meltsufin, line 103: `decrementRank()`  or `reduceRank(int)`

capstone/src/main/resources/static/map.js
jsheng112: Yeah I just realized this branch is behind

capstone/src/main/java/com/step902020/capstone/EventController.java
jsheng112: Yeah I tested this feature, the events all return in correct order on the backend. I think it has to do with the order that the divs are placed and the way they "float".

capstone/src/main/java/com/step902020/capstone/EventController.java
jsheng112: No its not. find by example does not support any other form of concatenated fields. I was't able to add "OrderByRankDesc"

capstone/src/main/resources/static/event.js
```diff
@@ -97,39 +97,42 @@ function selectedFilter(elementId){
...
  */
 function createEventElement(eventListElement, event, isIndividual, userEvent, userEmail) {
-  const eventElement = createElement(eventListElement, 'li', '');
-  eventElement.className = 'event';
-
-  // Click for event detail modal
-  eventElement.addEventListener('click', () => {
-    showEventPage(event, isIndividual, userEmail);
-  });
+  fetch("get-public-profile?organization-id=" + event.organizationId).then(response => response.json()).then((data) => {
```
lbourret, line 100: hm.. I think the problem with printing in rank order comes from the varying return times from fetch

capstone/src/main/resources/static/map.js
```diff
@@ -53,9 +66,11 @@ function createMarker(event,campusMap) {
  * @param campusMap - Google Map object of campus
  * return newInfoWindow = returns created info window
 */
-function createInfoWindow(event, campusMap) {
+function createInfoWindow(event, campusMap, isIndividual, userEmail) {
     var eventPosition = {lat: event.eventLatitude, lng: event.eventLongitude};
-    var eventContent = '<p id=mapContent>'+ event.eventTitle + '</p>';
+    var eventContent = document.createElement("div");
+    eventContent.className = "map-info-window";
+    createEventElement(eventContent, event, isIndividual, false, userEmail)
```
lbourret, line 73: nice!

capstone/src/main/java/com/step902020/capstone/EventController.java
meltsufin: Have you tried adding a `Sort` argument instead?
Something like:
```
Iterable<Event> findAll(Example<Event> example, Sort sort);
```

capstone/src/main/resources/static/event.js
jsheng112: Resolved

capstone/src/main/java/com/step902020/capstone/EventController.java
jsheng112: No I have not I can definitely try it.

PR #151, opened 2020-07-30
add functionality to display map of selected university

oasemota99:


capstone/src/main/resources/static/map.js
```diff
@@ -46,33 +70,53 @@ function createMarker(event,campusMap) {
   newMarker.addListener('click', function() {
     eventInfoWindow.open(campusMap, newMarker);
   });
+  var eventInfoWindow = createInfoWindow(event, campusMap);
```
jsheng112, line 73: Where is this info window used? No info window is showing up on the form.

oasemota99: should be fixed now!

PR #155, opened 2020-07-31
Add integration tests for recommendations

### LINKED ISSUE #147: add integration tests for recommendations

jsheng112:
- Split UserDatastoreTest into two files: IndividualControllerDatastoreTests and OrganizationControllerDatastoreTests
- Add recommender integration tests to IndividualControllerDatastoreTests (fixes #147)
- Add integration tests for calendar events
- Add integration tests for organization search
- Move review tests for organizations into OrganizationControllerDatastoreTests
- Fix minor map bug because this may be our last PR


capstone/src/test/java/com/step902020/capstone/UserDatastoreTest.java
```diff
@@ -248,6 +289,89 @@ public void testOrganizationDeleteEvent() throws URISyntaxException {
...
+  public void testGetIndividualRecommendedEventsOneRecommendedEventsReturned() throws URISyntaxException {
+    List<Individual> users = this.individualRepository.findByUniversity(expectedIndividual.getUniversity());
+    Recommender mockRecommender = Mockito.mock(Recommender.class);
+    List<Event> recommendedEvents = new ArrayList<Event>();
+    recommendedEvents.add(expectedEvent2);
+    when(mockRecommender.recommend(expectedIndividual, users, u -> u.getSavedEvents(), 2)).thenReturn(recommendedEvents);
+    IndividualController unitUnderTest = new IndividualController();
+    unitUnderTest.setRecommender(mockRecommender);
+    // getting the actual result
+    Event[] result = authRestTemplate.getForObject("/get-recommended-events-individual?count=2", Event[].class);
```
elefeint, line 317: How do you know these two recommended events contain one "recommended" and one "filler" event?

capstone/src/test/java/com/step902020/capstone/UserDatastoreTest.java
```diff
@@ -248,6 +289,89 @@ public void testOrganizationDeleteEvent() throws URISyntaxException {
...
+    assertEquals("Wrong number returned", 2, result.length);
```
meltsufin, line 318: To add to Elena's comment above, I would also check which events are returned rather than just the count, in order to have a more robust test.

jsheng112:
Right now the order that the events are added is more or less random (depending on the order that datastore returns them) which makes it hard to check the contents and compare them. I am going to wait until the rank PR gets merged before editing these tests.

capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -61,7 +61,7 @@
                 null, null, 0,
                 0, foodAvailable, requiredFee),
       ExampleMatcher.matching().withIgnorePaths("datastoreId", "organizationId", "eventLatitude", "eventLongitude", "rank")
-      ), Sort.by(Sort.Direction.DESC, "rank")
+      ), Sort.by(Sort.Direction.DESC, "rank_")
```
elefeint, line 64: I don't think you meant to commit this -- it was a test to make sure Datastore failed on a nonexistent column.

capstone/src/test/java/com/step902020/capstone/IndividualControllerDatastoreTests.java
```diff
@@ -0,0 +1,370 @@
...
+    when(mockRecommender.recommend(any(Individual.class), any(List.class), any(), same(2))).thenReturn(new ArrayList<Event>());
+    context.getBean(IndividualController.class).setRecommender(mockRecommender);
+    // getting the actual result
+    Event[] result = authRestTemplate.getForObject("/get-recommended-events-individual?count=2", Event[].class);
+    Event[] expected = new Event[2];
+    expected[0] = expectedEvent3;
+    expected[1] = expectedEvent2;
+    assertEquals("Wrong number returned", 2, result.length);
+    assertEquals("Wrong events returned at position " + 0, expected[0], result[0]);
+    assertEquals("Wrong events returned at position " + 1, expected[1], result[1]);
```
elefeint, line 256: Might as well make `0` and `1` part of the string; no need to concatenate.

capstone/src/main/java/com/step902020/capstone/Recommender.java
```diff
@@ -6,6 +6,7 @@
 import java.time.LocalDateTime;
 import java.util.*;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
```
meltsufin, line 9: There should be no `springframework` imports in this file at all.

jsheng112:
Sorry because of the power outage and the broken internet I was not able to push again

jsheng112:
I think something weird happened with my commits when I tried to push yesterday but the internet cut out. There are multiple merges of the branch into the branch and I tried to rebase but it did not work.

capstone/src/test/java/com/step902020/capstone/OrganizationControllerDatastoreTests.java
```diff
@@ -0,0 +1,152 @@
...
+    HttpEntity<MultiValueMap<String, Object>> request = new HttpEntity<>(map, headers);
+    ResponseEntity<String> response = authRestTemplate.postForEntity(url, request, String.class);
+    // getting the actual result
+    final String baseUrl = "/get-organization?email=" + expectedEmail;
+    URI uri = new URI(baseUrl);
+    Set<Long> expectedSet = new HashSet<>();
+    expectedSet.add(234L);
+    Organization result = authRestTemplate.getForObject(uri, Organization.class);
+
+    assertFalse("Unexpected organization ID",result == null);
```
meltsufin, line 129: There's `assertNull` which is more idiomatic here.

elefeint:
@meltsufin The null-failing tests are because the rest template is being autowired without base URL for some reason. I am looking into it.

elefeint:
@jsheng112 Well, this turned out to be simpler -- replace URL "user-info" with "/user-info", and the tests will pass.

I find this behavior odd.. but looking at Spring Boot code, it's very intentional -- https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RootUriTemplateHandler.java#L77

elefeint:
There is one test where there is still an issue (but a different one) -- EventDatastoreTest.testAddReview.testAddReview

jsheng112:
I think this might be a merge issue. Lian might have fixed it in another PR. I will merge master into my branch and run it again.

capstone/src/main/java/com/step902020/capstone/IndividualController.java
```diff
@@ -304,4 +309,58 @@ public String uploadImage(CurrentUser user) throws IOException{
...
+  }
+
+  /**
+   * Add new review to event
+   * @param user current user
+   * @param text review's text
+   * @param orgId organization's datastore id
+   * @return updated review list
+   */
+  @PostMapping("/add-org-review")
```
lbourret, line 328: Should this be in individual controller if it modifies the organization entity's data?

lbourret: Also, if we move functions into the controller of who triggers the function, would that mean we should move everything into the org or individual controllers? So add/remove/like/edit event reviews should also be in the individual controller and add/edit events should be in the organization controller?

jsheng112: Hmm, yeah that is true. The reason why I couldn't move it into the organizationtests was because to do these tests you need the current user to be an individual, but for the rest of the tests I need the current user to be an organization.
We can also move it back to individual, or maybe moving it into review tests?

PR #157, opened 2020-08-04
Readme

jsheng112:
- edit readme for this project and embed user instruction videos as gifs
- add javascript docs to all functions in user.js
- the gifs do not embed correctly in this branch because the gifs are not in the master branch yet. However I think the links will work fine once merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-# College Events Finder
+# Activity Hall
+
+The goal of this project is to create a platform that allows college students to find events that organizations they follow are currently hosting. The project allows users to find events on campus based on topics/organizations that they are interested in. This solves the frustration of searching for events on campus and includes monetary constraints to limit results to the users’ needs and by strengthening their connection to their community . This project is for everyone in college! Anyone who is looking for something to do on a boring saturday night by yourself or with friends will find this useful. We want to address the restrictions and considerations that different groups of people face, and make it so that these aspects are integrated into the search for university community-based events.
 
 ## Running Locally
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,3 +13,13 @@ mvn clean spring-boot:run -P local -DskipTests
...
 * Password: `test`
+
+## Running on App Engine
+
+This project is deployed at https://step90-2020.ue.r.appspot.com/ for a limited amount of time.
+
+## Using Activity Hall
+
+Please refer to these user guide videos to learn how to use our website:
+![](individualuser.webm)
```
meltsufin, line 24: If these embeds don't work in the README, which they probably don't, you can instead add them to you project web site ([GitHub Pages](https://pages.github.com/)), which I just generated in the [gh-pages](https://github.com/googleinterns/step90-2020/tree/gh-pages) branch.
See it live: http://googleinterns.github.io/step90-2020.
You can commit to the branch directly with no PRs, if you like.

capstone/src/main/resources/static/user.js
```diff
@@ -1,4 +1,7 @@
-/* helper function to highlight active tab for general nav */
+/**
+  * helper function to highlight active tab for general nav
```
elefeint, line 2: You can skip "helper function", and instead go with 3rd person -- JSdoc has fewer hard rules than Javadoc, but it's going to look familiar to other developers to see "Highlights active tab in general navbar".

As an example, here is Google JSdoc style guide: https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments

PR #161, opened 2020-08-06
add test for getting university

oasemota99:
small pr; I just wanted to get a couple of tests out

This tests the functionality to get the user's university based on their user type


capstone/src/test/java/com/step902020/capstone/EventDatastoreTest.java
```diff
@@ -131,7 +129,6 @@ public void testAddReview() throws URISyntaxException {
             expectedReview.text.equals(result.reviews.get(1).text));
   }
 
-  @Test
```
meltsufin, line 134: Is it because this test is failing?

PR #162, opened 2020-08-06
fix event campus map bug

jsheng112:
- fixes a small event map bug where the university was not fetched correctly. Before, the function getMapEvents() would only fetch correctly if the user was an individual. When the user is an organization, the fetching fails because the function is written with individualRepository
- fix a small calendar event issue where the save event button is displayed but it is confusing to the user because the user cannot tell whether this is actually one of their saved events or the event is from one of the saved organizations. To solve this, I removed all the save event buttons from the calendar page.

### DIFF
--- a/capstone/src/main/java/com/step902020/capstone/EventController.java
+++ b/capstone/src/main/java/com/step902020/capstone/EventController.java
@@ -89,8 +89,8 @@ public class EventController {
   }
 
   @GetMapping("get-map-events")
-  public Iterable<Event> getMapEvents(CurrentUser user) {
-    University university = this.individualRepository.findFirstByEmail(user.getEmail()).getUniversity();
+  public Iterable<Event> getMapEvents(CurrentUser user, @RequestParam("university") String universityName) {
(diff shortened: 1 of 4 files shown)


capstone/src/main/java/com/step902020/capstone/EventController.java
```diff
@@ -89,8 +89,8 @@
   }
 
   @GetMapping("get-map-events")
-  public Iterable<Event> getMapEvents(CurrentUser user) {
-    University university = this.individualRepository.findFirstByEmail(user.getEmail()).getUniversity();
```
meltsufin, line 93: I wonder why this doesn't work.

jsheng112: This function works if the user is an individual, but if the user is an organization it fails on the first line because it uses individualRepository.

meltsufin: I see. OK!

